PR #94, opened 2015-03-12
UF-173 : VersionRecords that have the same timestamp are in the wrong or...

Rikkola:
...der


Rikkola:
And now that I did the pull request. I can not get all the tests to pass. So hold the merge.

Rikkola:
Ok, the tests fail even without this commit. So should not be caused by this.

PR #95, opened 2015-03-17
Fix failing tests + make Git tests more robust

psiroky:
* the tests were failing just when executed directly
   using Maven (running from IDE was OK). However,
   the clean-up should be done anyway.
 * the Git tests are using random new port every
   time. This makes them more robust in case there
   is rogue process listeting on the default port

I think it makes sense to cherry-pick the fixes also into master and 0.6.x.

### DIFF
--- a/uberfire-nio2-backport/uberfire-nio2-impls/uberfire-nio2-jgit/src/test/java/org/uberfire/java/nio/fs/jgit/AbstractTestInfra.java
+++ b/uberfire-nio2-backport/uberfire-nio2-impls/uberfire-nio2-jgit/src/test/java/org/uberfire/java/nio/fs/jgit/AbstractTestInfra.java
@@ -22,6 +22,7 @@ import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
+import java.net.ServerSocket;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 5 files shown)


manstis:
@porcelli - would you mind taking a look please (since it relates to nio2 related things)

@psiroky - do you mean to target 0.5.x? This would affect only KIE 6.2.x code. We'd need the same on master to apply to KIE 6.3.0-SNAPSHOT.

psiroky:
@manstis agreed, we should fix this on master too (and probably also 0.6.x?). See the last line in the description.

uberfire-nio2-backport/uberfire-nio2-impls/uberfire-nio2-jgit/src/test/java/org/uberfire/java/nio/fs/jgit/AbstractTestInfra.java
```diff
@@ -124,4 +140,18 @@ public PersonIdent getAuthor() {
...
 
+    public static int findFreePort() {
+        int port = 0;
+        try {
+            ServerSocket server =
+                    new ServerSocket(0);
+            port = server.getLocalPort();
+            server.close();
+        } catch (IOException e) {
+            fail("Can't find free port!");
```
porcelli: fail method can't be resolved. I had to add this in order to compile:

import static org.junit.Assert.fail;

psiroky: Fixed. Thanks for pointing this out.

PR #97, opened 2015-03-19
BZ-1203696 - Access restrictions to assets does not work with custom rol...

mswiderski:
...es


porcelli:
this is a quick and dirty fix... this has to be improved.

mswiderski:
come on, it's not dirty it's just quick ;) + 1 needs to be improved

porcelli:
@mswiderski so... tell me about a RoleDecisionManager type that uses Roles and groups... if not dirty... what could be? ;)

mswiderski:
I usually call it "intermediate solution" but you have point with name suggests something else ;)

PR #98, opened 2015-03-20
BZ-1203696: clear separation between Roles and Groups

porcelli:


manstis:
http://github.com/uberfire/uberfire/commit/804042269

PR #100, opened 2015-04-08
BZ-1206627, BZ-1198662: improvements on shutdown and cluster state

porcelli:


uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/io/DisposableShutdownService.java
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        for ( final PriorityDisposable disposable : collection ) {
+            disposable.dispose();
+        }
+
+        SimpleAsyncExecutorService.shutdownInstances();
+
+        for ( final FileSystemProvider fileSystemProvider : FileSystemProviders.installedProviders() ) {
+            if ( fileSystemProvider instanceof Disposable ) {
```
manstis: I assume FileSystems need to be shut-down second from last (clsuter being the last) and hence you handle them here, after the other PriorityDisposables.. my question is: why not make FileSystemProviders (and SimpleAsyncExecutorService and ClusterService) also implement PriorityDisposable and have everything be disposed of using the same mechanism?

uberfire-commons/src/main/java/org/uberfire/commons/cluster/ClusterService.java
```diff
@@ -1,19 +1,17 @@
...
 
+import org.uberfire.commons.lifecycle.PriorityDisposable;
 import org.uberfire.commons.lock.LockService;
 import org.uberfire.commons.message.MessageHandlerResolver;
 import org.uberfire.commons.message.MessageService;
 
 public interface ClusterService extends MessageService,
-                                        LockService {
+                                        LockService,
+                                        PriorityDisposable {
```
manstis: See my comment in DisposableShutdownService... if ClusterService implements PriorityDisposable won't it be shut-down potentially twice? Once because it's a PriorityDisposable and again because DisposableShutdownService shuts down the cluster explicitly.

uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/io/DisposableShutdownService.java
porcelli: ClusterService and FileSystemProviders aren't managed beans.. so those can't be injected. Of course I could add everything to the ordered collection... but in this case I prefer to be more explicit.

uberfire-commons/src/main/java/org/uberfire/commons/cluster/ClusterService.java
porcelli: Not today as ClusterService is not a managed bean - we use ClusterServiceFactory and ClusterServiceFactoryProducer to instantiate it.

manstis:
http://github.com/uberfire/uberfire/commit/f55557b52

PR #103, opened 2015-04-24
Gwt upgrade n patter fly

porcelli:


manstis:
http://github.com/uberfire/uberfire/compare/ee520f9...70fffe1

PR #104, opened 2015-04-27
Remove 3rd party repositories

psiroky:
* all dependencies  should be available in Maven Central
   or JBoss Public repository. Tried with -Dfull and empty
   local maven repo and everything builds successfully.

 * We should not use any third party repos to ensure builds are reproducible, fast, etc. See https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/blob/master/README.md#requirements-for-dependencies for more details.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -74,20 +74,6 @@
         <updatePolicy>daily</updatePolicy>
       </snapshots>
     </repository>
-    <repository>
-      <id>sonatype</id>
-      <url>http://oss.sonatype.org/content/repositories/snapshots</url>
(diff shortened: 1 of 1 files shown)


manstis:
Does it still compile with these repositories removed?

psiroky:
Yes, tried with `-Dfull -Dmaven.repo.local=maven-repo` (the maven-repo dir was empty).

manstis:
http://github.com/uberfire/uberfire/commit/297999768

manstis:
Thanks Petr.

PR #106, opened 2015-04-29
fixing deployment and distro

porcelli:


uberfire-parent-with-dependencies/pom.xml
```diff
@@ -49,6 +52,10 @@
 
     <version.org.picketlink>2.6.0.CR2</version.org.picketlink>
     <version.org.jboss.logging>3.1.1.GA</version.org.jboss.logging>
+    <version.org.jboss.resteasy>2.3.7.Final</version.org.jboss.resteasy>
```
psiroky: Can we use 2.3.10.Final? To be in sync with IP BOM.

porcelli: I'll check from where this came... I think it's from Errai... anyway I'll check

uberfire-showcase/showcase-distribution-wars/src/main/tomcat7/WEB-INF/web.xml
```diff
@@ -5,7 +5,9 @@
 		  http://java.sun.com/xml/ns/javaee/web-app_2_5.xsd"
          version="2.5">
 
-  <distributable/>
+  <listener>
+    <listener-class>org.jboss.weld.environment.servlet.Listener</listener-class>
+  </listener>
```
psiroky: Shouldn't we register the RestEasy servlet somewhere here? Or is the RestEasy actually needed, since its not being registered here?

porcelli: Good point, although UF per se doesnt' use it. I'll double check with Errai (again).

uberfire-parent-with-dependencies/pom.xml
psiroky: It seems Errai is using the ip-bom (or at least ip-bom-parent), so the version there should be 2.3.10.Final.

manstis:
Is this "good to go" now?

porcelli:
i still need to clarify a couple of things with christian, i'll let you know as soon as I get back from him.

porcelli:
ready ;)

uberfire-showcase/showcase-distribution-wars/src/main/tomcat7/WEB-INF/web.xml
porcelli: No more resteasy on classpath!

manstis:
http://github.com/uberfire/uberfire/commit/a792fe58c

PR #110, opened 2015-05-19
BZ-1066962: post commit hook support added, but with limitation imposed

porcelli:
by jgit. ie. windows needs cygwin. more info:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=299315


manstis:
It looks like hooks only run if detectedFS has been set (to something other than the default, that I assume does a no-op?). How does detectedFS get set? I see in your tests you do it manually, but how would, for example, someone add hooks to kie-wb?

porcelli:
Here is setted by default:
https://github.com/uberfire/uberfire/pull/110/files#diff-78a9b405512867dba4d16bc7eb0e3cbcR217

User's can define a folder that, for every new repo we create... we'll copy the hooks from:

https://github.com/uberfire/uberfire/pull/110/files#diff-78a9b405512867dba4d16bc7eb0e3cbcR224

manstis:
Ok, I thought FS.DETECTED was a default implementation with a no-op
function. I'll look again tomorrow when at my PC.

> Sent when I shouldn't really be working.

porcelli:
Nope, FS.DETECTED is the detected File System by JGit.

(I just didn't use FS.DETECTED in order to be able to test it - you see: i'm a good boy)

manstis:
http://github.com/uberfire/uberfire/commit/e07ced2ae

I had to fix the test (as it assumed File[] was ordered the same on different JDK's): https://github.com/uberfire/uberfire/commit/e07ced2ae#diff-99c655c635f7a3453e21ccdd280c7befR83

I also had to add an additional "null" check: https://github.com/uberfire/uberfire/commit/e07ced2ae#diff-78a9b405512867dba4d16bc7eb0e3cbcR243

PR #112, opened 2015-05-27
UF testing tools project. First tool: a way to mock Event<Type> in cdi beans

ederign:


uberfire-testing-utils/src/main/java/org/uberfire/event/EventSourceMock.java
```diff
@@ -0,0 +1,24 @@
...
+
+    @Override
+    public Event<T> select( Annotation... qualifiers ) {
+        throw new RuntimeException( "mocking testing class" );
+    }
+
+    @Override
+    public <U extends T> Event<U> select( Class<U> subtype,
+                                          Annotation... qualifiers ) {
+        throw new RuntimeException( "mocking testing class" );
```
porcelli: UnsupportedOperationException ?

manstis:
It'll be interesting to see how the Observer methods can be implemented without CDI. Arguably testing the Observers of Events is more important (OK, OK, as important!) than checking the firing of Events.

ederign:
I agree, but we are talking here about unit testing. So in that case if you are testing class X and want to verify how it reacts by a event, just call the observer method in X.

But for integration tests is another topic ;)

uberfire-testing-utils/src/test/java/org/uberfire/mocks/EventSourceMockTest.java
```diff
@@ -0,0 +1,53 @@
...
+
+import com.google.gwtmockito.GwtMockitoTestRunner;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+
+import static org.mockito.Mockito.*;
+
+@RunWith(GwtMockitoTestRunner.class)
+public class EventSourceMockTest {
```
porcelli: duplicated?

ederign: Fixed.

porcelli:
https://github.com/uberfire/uberfire/commit/af077fcadd6f3bacd4fd0e636cae20309931429f

PR #114, opened 2015-05-29
Refactor EditorLockManagerImpl to use existing Notification mechanism's new 'singleton' event.

manstis:
This changes EditorLockManagerImpl to use the new 'singleton' Notification mechanism.

Tried and tested and works a treat. IMO it even looks nicer ;)


csadilek:
@manstis I thought we agreed last night to go with the UX team's (and my slight) preference to display the popup within the editor? Just talked to @mbarkley and his preference is in the in-editor popup as well. If we can make the existing notification appear within the editor, we get the best of both worlds, but until then I think the built-in notification (while prettier) is not better UX as less noticeable.

csadilek:
Having said that, it might be a good addition anyway to be able to specify where the built-in notification will appear. So, if we want to add that feature, we can switch to the built-in notification.

csadilek:
In case others want to look and vote. 

Here's the version using the built-in notification: https://www.dropbox.com/s/icd1b7akqpqcexw/lock-notification3.png

And here's the in-editor popup: https://www.dropbox.com/s/atortis0veovxea/lock-notification2.png

Seems pretty clear to me that the combination would be the best solution ;).

manstis:
@csadilek I've refactored NotificationManager to support "message lists" per "container". Ideally I'd like to have provided an "Activity" in the NotificationEvent; however the sub-interface that has a getWidget() is in a different module; depending on which creates a circular dependency.

Have a look. see what you think.

csadilek:
Merged! The only thing I added was the ability to make the initial vertical spacing configurable so the lock notifications line up nicer with the editors. Looks nice now. I think we have the best of both approaches now.

PR #122, opened 2015-06-05
Minor JavaDoc fix

yurloc:
Moving `@param clazz` element to the method the has the `clazz` param.


manstis:
http://github.com/uberfire/uberfire/commit/2465b2d1a

Thanks Jiri.

PR #127, opened 2015-06-12
[BZ-1207318] change package name for copied the Apache commons-io classes

psiroky:
* the classes are basically the same as in commons-io jar,
   so adding the commons-io dependency should be transparent
   for users of the nio2-model


porcelli:
@psiroky there's one problem... this is a client side enabled lib (compiled by GWT). In order to gwt compile it need access to code, so the change should be more profound than that =/

psiroky:
I see. I did build the uberfire, but without the -Dfull (which contains the GWT compilation where it probably fails). I think that in that case we should just rename the package to something like org.uberfire.common.io so that it does clashes with the commons-io jar.

psiroky:
@porcelli please take a look. The UF build was successful with this change. I also compiled all the KIE repos starting from guvnor up to the kie-wb-distributions -- no issue found.

porcelli:
https://github.com/uberfire/uberfire/commit/4c99673c612d71abbe38b5f048286e27bc5d729c

PR #128, opened 2015-06-15
UF-194: archetypes fixed

porcelli:


manstis:
Couple of problems, I'm afraid:

1) ```mvn gwt:run``` the generated ```-showcase``` webapp gives [this](https://gist.github.com/manstis/24481f2b7694e2bd771d).

2) Deployed EAP-6.4 WAR and it looks like an image is missing. See [here](http://pasteboard.co/1irsECUk.png).

porcelli:
Thx @manstis 

1) seems that you executed gwt:run in the wrong place.. you have to go to showcase/webapp
2) I'll check

manstis:
http://github.com/uberfire/uberfire/commit/11c0aed91

PR #131, opened 2015-06-25
archetype fix: image on login and gitignore on show case dir

ederign:


manstis:
http://github.com/uberfire/uberfire/commit/f7a2a3097

PR #132, opened 2015-06-25
Special Flow Panel (with height) required by tutorial and templated p…

ederign:
…erspectives. This is related with: https://issues.jboss.org/browse/UF-185. This panel used to live as default in our archetype


manstis:
http://github.com/uberfire/uberfire/commit/c40a0cee2

PR #136, opened 2015-07-02
UberFire new navigation system

cristianonicolai:
Navbar switch between compact and standard modes.


manstis:
Oee.... oee... how exciting is it on your showcase?

manstis:
I worry about a couple of things:

1) The impact PerspectiveMenuItem will have on all other Web apps (that currently use a regular MenuItem to navigate to different perspectives).

2) There is a UserPreferences service already in (either uf-ext or kie-wb-common; I forget which). Please, please, please let's keep a single service (in uf probably) and refactor the others. Speak to @salaboy or @neus (Walter would be interested too.... but I forget his nick).

cristianonicolai:
Hi @manstis! Very excited about this new feature too! :)

As for 1, there should be no impact on current menu items. They will continue to work as is. What the new PerspectiveMenuItem introduces is a way to track the current perspective and set is as "selected" in the navbar. Ideally, for the menu to always highlight the current perspective, menu items that represent a perspective change would have to be changed to PerspectiveMenuItem.

2 - Thanks for letting me know, no point on duplication. I checked that there is UserPreferences service on uf-ext already. So the question is, should that be move into the core? It does introduces a backend service whereas my approach was to use Cookies to store the data. I think that not only user preferences but global (app) preferences will be needed in the near future. Mark Procktor was mentioning about a plugin preferences view editor which will also require some sort of storage mechanism. If bringing those services into the core engine isnt a concern I think that would be my suggestion. If we decide to go ahead with it I can remove the userpreference from this commit and make that a separated PR. @porcelli any thoughts?

manstis:
Wouldn't it be great (IMO!) if:-

1) PlaceManager recognised a place request was for a perspective and then
could maintain the "current perspective" for the NavBar. Then you could
also launch perspectives from anywhere you can do "PlaceManager.goto(... )"
and not just Menu items.

2) UserPreferences could have pluggable/configurable persistence so some
preferences could be stored as cookies whilst others use a backend. This
would be the best of both worlds.. It should (IMO) live in UF and replace
the other UserPreferences in UF-EXT.

@porcelli... thoughts?

porcelli:
@cristianonicolai agree with @manstis - we need avoid this perspective menu... in fact we can keep it, but the "selecion" of the menu item should be the opposite (instead of selection comes from menu click, it should come from the perspective change - enabling the "goTo" as @manstis mentioned). In order to obtain perspective info you can use/adapt PerspectiveManager+PerspectiveChange event.

Regarding UserPreferences: let's for now use UserServicesBackendImpl and make it simple.

porcelli:
Regarding PerspectiveMenu: it would be good to add more flexibility to it, additional to "String identifier" also support a more flexible/generic PlaceRequest.

manstis:
In PlaceManger.goto(...) you could also check if the Activity is an
instance of AbstractWorkbenchPerspective (or whatever) to detect a
perspective has been navigated to.

cristianonicolai:
Guys, I reviewed the implementation for the user preferences. As suggested, Im now reusing the same UserPreferencesService and UserServicesBackendImpl that is implemented on uf-ext. This way, widgets are not tied to the particular storage implementation and neither services are duplicated. The only downside is that changes will be required on uf-ext and kie-wb-common (WorkbenchConfigurationPresenter) to get it working. The change for  uf-ext is already linked is here, for kie-wb-common I expect that to be done once the migration to PatternFly is done.

cristianonicolai:
As for the perspective menu, I'm a bit confused about that. You guys mentioned the ability to do PlaceManger.goto(...) and use PerspectiveChange events. I think this is what I got in here. WorkbenchMenuBarPresenter is tracking PerspectiveChange events and selecting the appropriate menu for that.  Also, the implementation of each MenuItemPerspective command is to actually do a PlaceManger.goto() as its click handler.
My idea with a custom type of menu is that somewhere/somehow I need to link a perspective to its menu from the navbar. Using the perspective identifier and the custom type solved the problem. If I use MenuItemCommand for instance, I'm not sure how I would be able to identify that in that command the is a perspective change to perspective x. Plus, it sounded to me a good way to enforce that this type of menu item will be used for this specific purpose (switching perspectives) and that is the only menu that gets tracked with an active state. 
Anyway, I might be missing something here, so feel free to shoot :)

manstis:
What if I want a Button in an editor to open a Perspective. No menu item
anywhere (I think Eder has this type of thing already - if you open a file
from the Social Activities Time Line it opens the editor in the Authoring
Perspective).

So by not relying on MenuItems to determine what the "active" perspective
is allows the "Active Perspective tracking" to work regardless of how the
Perspective was launched.

Using the type of Activity looked up in PlaceManager.goto(...) would mean
you could identify requests to change Perspective from anywhere (menus,
Buttons, etc).

porcelli:
@cristianonicolai as we talk today - base your last adjust on this: https://github.com/uberfire/uberfire/pull/140

cristianonicolai:
@manstis yep, got what you're saying. If in anywhere we've an action to switch the perspective, either a button or a menu item, you're proposing to keep the current perspective as a variable inside placemanager. I think that is fine but I don't really need that state for the navbar. The way I'm currently updating the navbar is by relying on the PerspectiveChange event. Every time a perspective is changed, could be from the navbar itself or from any other place, a button for instance, that event is fired with the new perspective, so I know which menu to highlight. The only purpose for the menu items in this case is to link a perspective to the widget that needs to be highlighted. WorkbenchMenuBarPresenter.onPerspectiveChange is where I tied it together.

Btw, I notice that we already have the current perspective state available on PerspectiveManager.getCurrentPerspective.

manstis:
@porcelli OK, so Cristiano's comment above, and out chat on IRC, has clarified things a lot for me. I misunderstood the usage/sequence ```MenuItemPerspective``` had.. believing clicking on the ```MenuItemPerspective``` caused the "active" Perspective to be recorded; however ```MenuItemPerspective``` is used by ```WorkbenchMenuBarPresenter``` to visually highlight a ```MenuItem``` based on the ```PerspectiveChangeEvent``. ```MenuItemPerspective``` is needed to record the Perspective's Identifer.

Happy to +1 this PR now.

porcelli:
https://github.com/uberfire/uberfire/commit/1c5ad4e66

cristianonicolai:
@porcelli I think the wrong commit got merged. My latest update did not included the UserPreferences class as we discuss. I guess it could be because I updated this PR with a new commit, see https://github.com/cristianonicolai/uberfire/commit/11a5e2dbcfc20a6821f9bb9e0f238a99a982cabb.

porcelli:
fixed https://github.com/uberfire/uberfire/commit/11629fbed0e2a306770d2bb453185be924326048

PR #137, opened 2015-07-03
Included methods on multi page editor to keep compatibility with prev…

cristianonicolai:
…ious implementation from uf-extensions


porcelli:
http://github.com/uberfire/uberfire/commit/97333e441

PR #139, opened 2015-07-08
Errai version update

cristianonicolai:


porcelli:
We agreed to keep using SNAPSHOTS for Errai as we're in the same team

PR #144, opened 2015-07-21
UF Docks support

ederign:


porcelli:
https://github.com/uberfire/uberfire/commit/04f7513cd

PR #150, opened 2015-08-07
File System providers resource Fix

ederign:
Requested by @qmx.

FileSystemTestingUtils utils requires git file system providers declared in resources.


qmx:
yay, it works!

porcelli:
https://github.com/uberfire/uberfire/commit/81d62d63d

PR #158, opened 2015-08-21
Screen context buttons on uberfire docks

ederign:


ederign:
Merged https://github.com/uberfire/uberfire/commit/4916d00a5

PR #163, opened 2015-08-31
New widgets to support PatternFly Field Level Help style.

cristianonicolai:


uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/widgets/FormLabelHelp.java
```diff
@@ -0,0 +1,70 @@
+/*
+ *
+ *  * Copyright 2012 JBoss Inc
```
psiroky: It is 2015 :)

porcelli:
http://github.com/uberfire/uberfire/commit/09a1890db

Just note, this is the component: https://www.patternfly.org/patterns/field-level-help/

PR #166, opened 2015-09-04
Place Hid event on Uberfire

ederign:
When a editor is hid in ListBarWidgetImpl, we would like to generate a event (PlaceHidEvent). This should work like Focus/Maximized events.

We have to change ActivityManagerImpl, because when the part is associated with PathPlaceRequest, the place identifier was null (making impossible to
identify which place was hidden).

So in order to do that, when an activity is resolved by ActivityManagerImpl, I've made a lookup on resolvePlaceIdentifierForPathPlaceRequest method.

This feature request was generated by the need for hide the docks when Data Modeller was put in the background.

### DIFF
--- a/uberfire-workbench/uberfire-workbench-client-views-bs2/src/main/java/org/uberfire/client/views/bs2/listbar/ListBarWidgetImpl.java
+++ b/uberfire-workbench/uberfire-workbench-client-views-bs2/src/main/java/org/uberfire/client/views/bs2/listbar/ListBarWidgetImpl.java
@@ -369,6 +369,7 @@ public class ListBarWidgetImpl
                 return true;
             }
             parts.add( currentPart.getK1() );
+            panelManager.onPartHid(currentPart.getK1());
             currentPart.getK2().getElement().getStyle().setDisplay( NONE );
         }
 
--- a/uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/workbench/PanelManagerImpl.java
+++ b/uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/workbench/PanelManagerImpl.java
@@ -1,17 +1,5 @@
 package org.uberfire.client.workbench;
 
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import javax.annotation.PostConstruct;
-import javax.annotation.PreDestroy;
-import javax.enterprise.context.ApplicationScoped;
-import javax.enterprise.event.Event;
-import javax.enterprise.event.Observes;
-import javax.enterprise.inject.Instance;
-import javax.inject.Inject;
-
 import com.google.gwt.core.client.Scheduler;
 import com.google.gwt.core.client.Scheduler.ScheduledCommand;
 import com.google.gwt.event.logical.shared.AttachEvent;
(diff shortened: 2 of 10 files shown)


manstis:
License headers and tests please!

manstis:
Just a thought whether it'd be better to have PathPlaceRequest use IOC to get a reference to ActivityManager and resolve an identifier in the constructor (like we do for ObservablePath)? Otherwise the identifier is not set until a request to go to a PathPlaceRequest is handled by PlaceManager (and consequentially ActivityManager). I just think there's a greater risk of something else relying upon a PathPlaceRequest identifier in the future that doesn't pass through ActivityManager... IDK what ;-)

ederign:
@manstis the news classes has license headers (PlaceHidEvent). Should I miss something?

About testing, I did not write tests for this new feature because ActivityManagerImpl nas no  unit test on it :( . 

I think we should cover everything in another task what do you think? (I can put that in my task list).

manstis:
Hi Eder... I deleted my comment about missing license headers (forgetting I
was just looking at diffs; doh!) As for tests.... well I never thought I'd
hear you say that!!! Shocker! We should probably try to write something at
some time.

ederign:
@manstis I fell bad about it, but it's a product bug. With pain in my heart I accept to leave the tests for later.



https://www.google.com.br/url?sa=i&rct=j&q=&esrc=s&source=images&cd=&ved=0CAcQjRxqFQoTCOz1wdyX3scCFZGyHgodPDII3g&url=http%3A%2F%2Fwww.buzzfeed.com%2Fchelseamarshall%2Fpugs-feeling-so-left-out&psig=AFQjCNGi_LwkKE1y4rPsUqLdtUWMDRCDLw&ust=1441483787608271

uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/workbench/PanelManagerImpl.java
```diff
@@ -290,6 +286,11 @@ public void onPartMinimized( final PartDefinition part ) {
     }
 
     @Override
+    public void onPartHid( final PartDefinition part ) {
+        placeHidEvent.fire( new PlaceHidEvent( part.getPlace() ) );
```
ederign: @wmedvede [2]

uberfire-workbench/uberfire-workbench-client-views-bs2/src/main/java/org/uberfire/client/views/bs2/listbar/ListBarWidgetImpl.java
```diff
@@ -369,6 +369,7 @@ public boolean selectPart( final PartDefinition part ) {
                 return true;
             }
             parts.add( currentPart.getK1() );
+            panelManager.onPartHid(currentPart.getK1());
```
ederign: @wmedvede [1]

manstis:
*uberfire*
(master) http://github.com/uberfire/uberfire/commit/fc41dbe45

(0.7.x) http://github.com/uberfire/uberfire/commit/3a63755fc

manstis:
Thanks @ederign - BTW, I removed the ```ActivityManagerImplTest``` class (see comment above).

PR #167, opened 2015-09-09
[BZ-1259233] download patternfly as maven artifact

psiroky:
* the Maven artifact was created using http://www.webjars.org/npm
   from the NPM package. It now resides in Maven central. 
   Basically for any new PatternFly version we would need to create the jar
   (or figure out how to do it automatically during PatternFly release)

 * the content is under META-INF/resources, but it can be easily
   extracted from there into the structure we need

@porcelli @cristianonicolai please take a look and let me know what you think


cristianonicolai:
Hi @psiroky, looks good to me. This is something I was actually thinking of doing anyway, thanks :)

psiroky:
@porcelli, @cristianonicolai so are you guys ok with merging this change?

porcelli:
https://github.com/uberfire/uberfire/commit/0267f8c7b

PR #168, opened 2015-09-09
[BZ-1259233] use version property for CDI 1.0 compatible Errai

psiroky:


porcelli:
https://github.com/uberfire/uberfire/commit/14f8a2343

PR #170, opened 2015-09-11
DataGrid PatternFly widget

cristianonicolai:


porcelli:
https://github.com/uberfire/uberfire/commit/8635fa0f807e2812773674ede09ed9ec368d9e11

manstis:
This is an interesting one.... uberfire-extensions (IIRC) has a SimpleTable and PagedTable widgets (I think extending GWT's DataGrid) that are used quite extensively by Drools and jBPM related screens... Does the BS3 upgrade change the super-type of the foregoing to this DataGrid so we'll all be in a happy place?

cristianonicolai:
@manstis thanks for the heads up. I've updated the SimpleTable in uf-ext so it also inherits from the common PatternFly DataGrid, avoiding the same fix and styling to be repeated over and over. Please check https://github.com/uberfire/uberfire-extensions/pull/65 and https://github.com/uberfire/uberfire/pull/171.

PR #175, opened 2015-09-15
WAS problem with UF background tasks like lucene indexes due to wrong

porcelli:
component being used - does not use EJB for async operations but simple
thread pool and thus has no access to CDI beans, fails with:
java.lang.ClassCastException:
com.ibm.ejs.container.deploy.java.util.concurrent.EJSLocal0SGSimpleAsyncExecutorService_ef40626b
incompatible with org.uberfire.commons.async.SimpleAsyncExecutorService

Revert "by implement an interface, it makes easy to mock"

This reverts commit bae32af57249f244937718b40fa67b58b039494f.


manstis:
*uberfire*
(master) http://github.com/uberfire/uberfire/commit/9dc372239

(0.7.x) http://github.com/uberfire/uberfire/commit/b76731bc6

manstis:
Thanks @porcelli !

PR #178, opened 2015-09-16
UF-209 - After all parts of a Panel are closed, it's not possible to …

cristianonicolai:
…open any other Part again


porcelli:
http://github.com/uberfire/uberfire/commit/0be162e26

PR #180, opened 2015-09-22
UF-225 Skipping unlock when RuntimeException is thrown in AbstractIOService.endBatch()

ederign:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1251358

Description of problem:

AbstractIOService.java
======================

        public void endBatch() {
            if ( !batchLockControl.isLocked() ) {
                throw new RuntimeException( "There is no batch process." );
            }

            if ( batchLockControl.getHoldCount() > 1 ) {
                batchLockControl.unlock();
                return;
            }

            if ( !fileSystems.isEmpty() ) {
                cleanupClosedFileSystems();
            }

            for ( final FileSystem fs : fileSystems ) {
                unsetBatchModeOn( fs );
            }

            batchLockControl.unlock();
        }  

If an RuntimeException is thrown before "batchLockControl.unlock()", the lock is not released so subsequent threads will wait for the lock forever. 

(for example)

    RULE throwEx
    CLASS org.uberfire.io.impl.AbstractIOService
    METHOD unsetBatchModeOn
    AT ENTRY
    IF TRUE
      DO throw new RuntimeException("dummy");
    ENDRULE

Solution:
===========

In  order to reproduce this, I've created a new test and a fix for it.

However, even if the lock is released the exception of the FS would leave the UF in an inconsistent state. Anyway, this PR raises an exception if that happens (and releases the lock).

### DIFF
--- a/uberfire-io/src/main/java/org/uberfire/io/impl/AbstractIOService.java
+++ b/uberfire-io/src/main/java/org/uberfire/io/impl/AbstractIOService.java
@@ -162,6 +162,18 @@ public abstract class AbstractIOService implements IOServiceIdentifiable,
             return;
         }
 
+        try {
+            cleanUpAndUnsetBatchModeOnFileSystems();
+        }
+        catch ( Exception e ){
+            throw new RuntimeException( "Exception cleaning and unsetting batch mode on FS." );
+        }
+        finally {
+            batchLockControl.unlock();
+        }
+    }
+
+    private void cleanUpAndUnsetBatchModeOnFileSystems() {
         if ( !fileSystems.isEmpty() ) {
             cleanupClosedFileSystems();
         }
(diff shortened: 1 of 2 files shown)


csadilek:
ok to test

ederign:
'ok to test'

csadilek:
master: 00bde1a5e43f9dd19d617d87bebb5b448aee9244
0.7.x: ca8c804a93cae2b619c0d2ca4b35169977dfeb7d

psiroky:
@csadilek, @ederign could you please try again commenting with the trigger phrase (without the quotes)? There was a configuration error in the Jenkins job, which should be now hopefully fixed.

psiroky:
Ah, I see the the PR was just merged. So let's test this with the next PR :)

ederign:
ok to test

psiroky:
Closed PRs are ignored, so the phrase will not work here anymore. After the fix, everyone with public membership (can be configured per user) in uberfire or droolsjbpm github org. units should be able to approve the PR + the build should be automatically triggered for them (without the need for the phrase).

mrietveld:
+100 -- this problem has been around for a while. Thank you for fixing this Eder!

PR #181, opened 2015-09-23
UF-226: restore the use of Executor interface in order to work on

porcelli:
WF/EAP. A improved solution still needed in the upcoming days that
would make it work not only on EAP/WF but also on WAS.
So it's clear that this quick fix will break UF on WAS for now.


manstis:
*uberfire*
(master) https://github.com/uberfire/uberfire/commit/5330357d2

(0.7.x) https://github.com/uberfire/uberfire/commit/e41913843

manstis:
Thanks @porcelli !

PR #184, opened 2015-09-24
[RFC] initial preferences API

qmx:


uberfire-api/src/main/java/org/uberfire/preferences/impl/DefaultResolutionStrategy.java
```diff
@@ -0,0 +1,33 @@
...
+import org.uberfire.preferences.ResolutionStrategy;
+import org.uberfire.preferences.Scope;
+
+@ApplicationScoped
+public class DefaultResolutionStrategy implements ResolutionStrategy {
+
+    @Override
+    public List<Scope> order( final Scope scope,
+                              final Scope... scopes ) {
+        return null;
```
porcelli: ?

porcelli:
Missing a reasonable test suite.
Also missing the client side bits.. that i can help on it... 30 mins is done.

porcelli:
Conversion tool from legacy to new format (now including a version info please).. would be very useful ;)

uberfire-api/src/main/java/org/uberfire/preferences/AsyncDictionary.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+package org.uberfire.preferences;
+
+import org.uberfire.mvp.ParameterizedCommand;
+
+public interface AsyncDictionary {
+
+    <T> void get( final String key,
+                  final Class<T> clazz,
```
csadilek: The parameter clazz doesn't seem to be used in the two implementations below.

csadilek:
Do we have a JIRA for this? If not, let's create one and reference it in the commit message, please. Just picking some low-hanging fruit....been a long day here ;).

uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/preferences/PreferenceStorageServiceBackendImpl.java
```diff
@@ -0,0 +1,138 @@
...
+    @Override
+    public void delete( final String key ) {
+
+    }
+
+    private Path buildStoragePath( final Scope scope,
+                                   final String key ) {
+
+        final String path;
+        if ( scope.equals( DefaultScopeTypes.USER.toScope() ) ) {
```
manstis: Would it not be better to ask the ```Scope``` to build a "prefix"? e.g. ```path = scope.getPrefix()```

uberfire-api/src/main/java/org/uberfire/preferences/impl/PreferenceStoreImpl.java
```diff
@@ -0,0 +1,85 @@
...
+        return defaultScope;
+    }
+
+    @Override
+    public ResolutionStrategy resolutionStrategy() {
+        return resolutionStrategy;
+    }
+
+    @Override
+    public void put( String key,
```
manstis: How would we put/get preferences to/from different scopes? It looks like we'd always be using ```defaultScope```?

uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/preferences/PreferenceStorageServiceBackendImpl.java
```diff
@@ -0,0 +1,138 @@
...
+    @Named("systemFS")
+    private FileSystem fileSystem;
+
+    private final XStream xs = new XStream();
+
+    @Override
+    public <T> T read( final Scope scope,
+                       final String key,
+                       final ResolutionStrategy resolutionStrategy ) {
+        List<Scope> scopes = resolutionStrategy.order( scope );
```
manstis: Based on @porcelli comment for ```DefaultResolutionStrategy``` and the ```@Produces``` you have in ```ApplicationScopedProducer``` won't this throw NPEs?

uberfire-api/src/main/java/org/uberfire/preferences/AsyncDictionary.java
porcelli: true that. the reason for that is basically that the default portable impl can't use clazz.cast as it won't compile on GWT. It should have a server side extension of it to take advantage of it.

uberfire-api/src/main/java/org/uberfire/preferences/impl/PreferenceStoreImpl.java
porcelli: AFAIK you'd use a forScope and from there you get what you're looking for. AND you'll also have the resolution strategy to define precedence.

uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/preferences/PreferenceStorageServiceBackendImpl.java
porcelli: this is what scope.key is for. the only exception is User scope as it has a special behavior (global state)

uberfire-api/src/main/java/org/uberfire/preferences/AsyncDictionary.java
csadilek: ok, then I guess it should be factored out on the server, but I think that's what you meant.

csadilek:
@qmx Can this (RFC) PR be closed assuming you will be sending a separate one for the feature?

manstis:
What's happening with a consolidated preferences service now? Is it dead?
It'd be a great shame - we really need one in our platform.

csadilek:
I agree with @manstis. Would really be a much needed feature.

PR #185, opened 2015-09-28
[UF-229] Configurable Workbench: Dynamic Screens are not proper displayed in Editor Screen on LayoutEditor and new Activity getIdentifier() method.

ederign:
This PR and the related [PR on uberfire-extensions](https://github.com/uberfire/uberfire-extensions/pull/79) group three changes:

1-) In order to prevent this (i.e. ActivityManagerImpl):

    private void resolvePathPlaceRequestIdentifier( PlaceRequest placeRequest, Activity activity ) {
        Set<Annotation> annotations = lookupActivityAnnotations( activity );
        for ( Annotation annotation : annotations ) {
            if ( annotation instanceof Named ) {
                String resolvedPlaceIdentifier = ( (Named) annotation ).value();
                placeRequest.setIdentifier( resolvedPlaceIdentifier );
                break;
            }
        }
    }

I've created a new method called String getIdentifier() on Activity.java

This requires a lot of changes in UF in order to support and generate this method.

Also, to solve:

2-) Dynamic screens are without the name in edit screen (they are blank) (after reload)
3-) When new static perspective plugin is created and browser is refreshed, it is not displayed in perspective plugin list.

I've created a class called ActivityBeansInfo, to list information about all beans available on class path and remove duplicated code on  [Edit Screens](https://github.com/uberfire/uberfire-extensions/blob/be19ebb86ece6016e368482519eeee20ae77cea3/uberfire-runtime-plugins/uberfire-runtime-plugins-client/src/main/java/org/uberfire/ext/plugin/client/perspective/editor/layout/editor/popups/EditScreen.java#L259-L275) and in [PluginNavList](https://github.com/uberfire/uberfire-extensions/blob/087db6a447ae4f90a21b0df4453b9557e6c01447/uberfire-runtime-plugins/uberfire-runtime-plugins-client/src/main/java/org/uberfire/ext/plugin/client/widget/navigator/PluginNavList.java#L147-L165) and [here](https://github.com/uberfire/uberfire-extensions/blob/087db6a447ae4f90a21b0df4453b9557e6c01447/uberfire-runtime-plugins/uberfire-runtime-plugins-client/src/main/java/org/uberfire/ext/plugin/client/widget/navigator/PluginNavList.java#L237-L244).

In both classes there was a bug where if the screen was generated via dynamic screens plugin (static or created on runtime plugins editor) the name is blank because there is no annotation associated with that.

One thing to keep in mind (and maybe we can discuss about this evolution in future) is that there is no way to use ActivityManager to list all avaliable screens because ActivityManager only "activate" a screen after a goTo to this activity (lazy one). So in order to list all the avaliable screens we have to search that dinammically via bean manager.

### DIFF
--- a/uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/mvp/ActivityBeansInfo.java
+++ b/uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/mvp/ActivityBeansInfo.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2015 JBoss Inc
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.uberfire.client.mvp;
+
+import org.jboss.errai.ioc.client.container.IOC;
+import org.jboss.errai.ioc.client.container.IOCBeanDef;
+import org.jboss.errai.ioc.client.container.SyncBeanManager;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Named;
+import java.lang.annotation.Annotation;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+
+
+@ApplicationScoped
+public class ActivityBeansInfo {
+
+    public List<String> getAvailableWorkbenchScreensIds() {
+        return lookupBeansId( WorkbenchScreenActivity.class );
+    }
+
+    public List<String> getAvailablePerspectivesIds() {
+        return lookupBeansId( PerspectiveActivity.class );
+    }
+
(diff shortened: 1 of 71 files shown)


uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/mvp/ActivityBeansInfo.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    public List<String> getAvaliableSplashScreensIds() {
+        return lookupBeansId( SplashScreenActivity.class );
+    }
+
+    public List<String> getAvaliableWorkbenchEditorsIds() {
+        return lookupBeansId( WorkbenchEditorActivity.class );
+    }
+
```
csadilek: Available is misspelled in the four methods above. Since it's public API we should prob. fix that.

csadilek:
Looks good! I think this is ready to merge after fixing the method names in ActivityBeansInfo.

uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/mvp/ActivityBeansInfo.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+
+
+@ApplicationScoped
+public class ActivityBeansInfo {
+
+    public List<String> getAvailableWorkbenchScreensIds() {
```
ederign: @csadilek fixed. Thanks for the review!

csadilek:
Thanks @ederign. 

Master: 2d1a5ca364f4b74a6bb498527040cfe2da560b9e

PR #186, opened 2015-09-29
Add no argument constructors to client-side application-scoped beans …

mbarkley:
This change helps prepare Uberfire for Errai 4.0, where normal scope beans will have to be proxiable.


mbarkley:
I will be attempting changes in Errai 4 so that no-argument constructors are no longer necessary for proxying. See discussion [here](https://github.com/uberfire/uberfire-extensions/pull/80) for details.

PR #188, opened 2015-10-14
UF-211 - Removed context dropdown

cristianonicolai:
Close and maximize/minimize buttons are now visible inline as other buttons inside toolbar.


ederign:
Merged: https://github.com/uberfire/uberfire/commit/2ced4ab89d9dc410807208765c4dc770a41688c5

Thanks!

PR #190, opened 2015-10-14
[from porcelli, leave here for now please] Git Squash operation

Salaboy:
Initial implementation for review.
@porcelli I didn't want to go deeper in the testing before checking the initial approach.

### DIFF
--- a/uberfire-nio2-backport/uberfire-nio2-impls/uberfire-nio2-jgit/src/main/java/org/uberfire/java/nio/fs/jgit/util/JGitUtil.java
+++ b/uberfire-nio2-backport/uberfire-nio2-impls/uberfire-nio2-jgit/src/main/java/org/uberfire/java/nio/fs/jgit/util/JGitUtil.java
@@ -84,8 +84,11 @@ import org.uberfire.java.nio.fs.jgit.JGitFileSystem;
 
 import static java.util.Collections.*;
 import static org.apache.commons.io.FileUtils.*;
+import org.eclipse.jgit.errors.IncorrectObjectTypeException;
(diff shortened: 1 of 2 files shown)


ederign:
@Salaboy IMO, we should not create PR's just for initial implementations. 

That only creates noise and you can have this conversation sending a PR to @porcelli  forks.

psiroky:
I agree with what @ederign. We should only create PRs against "blessed" repositories where the code is ready for review and merge.

manstis:
I also agree with both Eden and Petr.

Salaboy:
@ederign @psiroky @manstis I got the feeling that the issue was the word "initial" because the operation is complete and it is working. Do you see any other issue with the content of the PR? I think that the code is ready for review and merge, I was looking for that kind of feedback. I will avoid the word "initial" in my future PRs.

psiroky:
@Salaboy Yes, you are right. For me the word "initial" looked like something that is not yet ready.

Also when mentioning "I didn't want to go deeper in the testing...", does that mean more tests are on the way?

Salaboy:
@psiroky they will come in the future if they are needed, now the test provided check that the behaviour is the expected one

csadilek:
@porcelli is this one good to go?

porcelli:
@csadilek still on low priority, after BZs i'll be back to it

ederign:
@porcelli do you still need this PR? I would like to keep this list clean! :)

porcelli:
Yes, keep it there for now please

porcelli:
@adrielparedes is working on it, and will submit a new PR soon. finally we can close this ;)

PR #191, opened 2015-10-14
UF-237: SessionListener for lock cleanup not working on Tomcat7, E(J)WS

csadilek:
The problem was that dependency injection in @WebListeners
(HttpSessionListeners) is not supported in various containers
(some versions of Tomcat7, E(J)WS and WLS). Now our listener
is no longer relying on @Inject to get to its dependencies.

Luckily all its dependencies are already available after startup so
this was a relatively easy workaround.

This also includes test coverage.

**Needs to be cherry-picked to 0.7.x**


manstis:
*uberfire*
(master) http://github.com/uberfire/uberfire/commit/afbf901c7

(0.7.x) http://github.com/uberfire/uberfire/commit/80350ea32

manstis:
Thanks @csadilek !

PR #193, opened 2015-10-16
UF-236 - New ClosableSimpleWorkbenchPanelPresenter Panel Type

cristianonicolai:
New presenter type that supports closing current part, including tests as well as updated dashboard perspective to use presenter on showcase app.


csadilek:
Thanks, @cristianonicolai.

master: dcbda44a85ac894c9413631ad8243bba35f165ab

PR #194, opened 2015-10-19
UF241: Docks not hiding when different editor selected

ederign:
When switching between files, the newly introduced PlaceHidEvent is not fired, causing docks to stay active when they shouldn't.

Also renamed PlaceHidEvent to PlaceHiddenEvent.

Note, this only affects 0.8.0-SNAPSHOT. 0.7.0 works fine.

https://issues.jboss.org/browse/UF-241

### DIFF
--- a/uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/listbar/ListBarWidgetImpl.java
+++ b/uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/listbar/ListBarWidgetImpl.java
@@ -145,7 +145,7 @@ public class ListBarWidgetImpl
 
     WorkbenchPanelPresenter presenter;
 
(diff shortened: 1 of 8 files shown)


uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/listbar/ListBarWidgetImpl.java
```diff
@@ -300,6 +300,7 @@ public boolean selectPart( final PartDefinition part ) {
                 return true;
             }
             parts.add( currentPart.getK1() );
+            panelManager.onPartHidden( currentPart.getK1() );
```
ederign: that is the change.

ederign:
Child commit: https://github.com/droolsjbpm/kie-wb-common/pull/114

csadilek:
Thanks!

master: f43c2114af41646e584dba05ae1b92487d8cbd5f

PR #197, opened 2015-10-20
BZ-1202926 - Uploaded file name with spaces breaks upload, download, rename and copy operations

wmedvede:


uberfire-server/src/test/java/org/uberfire/server/FileDownloadServletTest.java
```diff
@@ -0,0 +1,287 @@
...
+        @Override public void flush() throws IOException {
+            output.flush();
+        }
+
+        @Override public void close() throws IOException {
+            output.close();
+        }
+    }
+
+    private class MockPath implements Path {
```
manstis: Did you really have to write a ```MockPath``` class, as opposed to just ```mock(Path.class)```?

No problem for this PR.. just a question.

manstis:
It wouldn't hurt to run "format code" over these classes; as there's a few formatting issues (but nothing significant to the PR really).

wmedvede:
It's true, the PathMock is not needed and it was removed.

csadilek:
@manstis is this a +1 from you? @wmedvede shouldn't this be applied to both master and 0.7.x?

manstis:
Not yet; but likely to be when I look at the PR again tomorrow. I won't
leave PRs ambiguous - they'll be clearly +1'ed when ready for merge.

manstis:
+1 to merge.

csadilek:
@manstis @wmedvede This should be on master too, right? If so, I can try cherry-picking and see if it merges without problem.

manstis:
@csadilek Yes please (or do you want separate PRs for other branches?)

csadilek:
@manstis ok thanks for confirming. I will do that later today. I will cherry-pick to both branches. For the future it was suggested to send separate PRs but we can discuss.

csadilek:
Thanks!

master: cefe42ce6b2837e7ecdaad197e07d6894d1ccf51
0.7.x: ff8a09d7efcf4e80c30426db1ee9333040f08a79

wmedvede:
My Understanding is that

1) an initial PR should be done, for example for 0.7.x
2) this initial PR is checked, bla, bla, bla, and finally merged into the
uberfire/0.7.x by a gatekeeper
3) After this, the original developer can generate a second PR for master
(or whatever any other required branches)

In my opinion this is good to avoid replicating all the eventually
modifications and bureaucracy on both PRs at the same time. (if both are
opened)
When the second PR is created it's expected that the reviewer only have to
do a quick check since all needed modifications has already been done in
the first PR









2015-10-20 22:24 GMT+02:00 Christian Sadilek <notifications@github.com>:

wmedvede:
see
https://docs.google.com/a/redhat.com/document/d/1NMY-j5Zkrnau3_PbcIfR5SJ4Fe8gxNVUvxXQI35z88I



   1.

      . It may often be beneficial to start with PR for master branch and
      after that one is reviewed and merged, create PRs for other
branches. This
      would eliminate the need to do same fixes on multiple branches.
      1.


2015-10-20 22:24 GMT+02:00 Christian Sadilek <notifications@github.com>:

csadilek:
True, @wmedvede. We're half way into implementing the process, I guess. I think the PR author should indicate though that there will be a follow-up PR to a different branch once merged. So, that the reviewers/gate keepers are aware.

PR #198, opened 2015-10-20
UF-177: +Use batch mechanism for locks to enforce cluster replication

csadilek:
startBatch/endBatch need to be called even when writing single files
to ensure cluster replication takes place. This needs to be addressed
in a future version of UF: https://issues.jboss.org/browse/UF-242

This is out of scope for the upcoming release so I added these calls
to the lock service.

**This is needed on both master and 0.7.x**


ederign:
Merged: master https://github.com/uberfire/uberfire/commit/d126ee7a8036ee159a303477fcfd7d560a0e01b3 , 0.7 https://github.com/uberfire/uberfire/commit/9dacd6b7c0c9cfd1a9c53bc3df84e552b150ad58

Thanks

PR #202, opened 2015-10-22
BZ:1271532: BPM cluster fails to setup due to NPE in uberfire

ederign:
Having a BPM cluster with 2 EAP nodes in a domain, the cluster fails to initialize due to NPE in org.uberfire.io.impl.cluster.helix.ClusterServiceHelix.getNodeStatus(). The first node starts up just fine, but the second node fails to deploy Business Central.

The issue occurs when cluster starts with a clean environment - including DB, .niogit and .index for both nodes,  due to nullview or empty map returned by Helix.
      
        String getNodeStatus() {
        final String partition = resourceName + "_0";
        final ExternalView view = getResourceExternalView();
        if ( clusterIsNotSetYet( view, partition ) ) {
            return "OFFLINE";
        }
        final Map<String, String> stateMap = view.getStateMap( partition );
        return stateMap.get( instanceName );
       }

      private boolean clusterIsNotSetYet( ExternalView view, String partition ) {
        //first start with fresh setup
        if( view==null ){
            return true;
        }
        final Map<String, String> stateMap = view.getStateMap( partition );

        return stateMap == null || stateMap.get( instanceName ) == null;
    }

       
and then the NPE used to happen here : view.getStateMap( partition ) or  stateMap.get( instanceName )

### DIFF
--- a/uberfire-io/src/main/java/org/uberfire/io/impl/cluster/helix/ClusterServiceHelix.java
+++ b/uberfire-io/src/main/java/org/uberfire/io/impl/cluster/helix/ClusterServiceHelix.java
@@ -1,11 +1,5 @@
 package org.uberfire.io.impl.cluster.helix;
 
-import java.util.HashMap;
-import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.atomic.AtomicInteger;
-import java.util.concurrent.locks.ReentrantLock;
-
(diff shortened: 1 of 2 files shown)


csadilek:
master: 763e5305639ff55049aba0eb4f1b5c0decd285d7
0.7.x: 9553004ed55b26c3fff5e823bc4247cbacdd33f3

PR #203, opened 2015-10-23
UF-177: +Improvements to cluster-replication for locks

csadilek:
Got time today to take a closer look how our IO clustering works 
and sending another update to improve cluster-replication for locks.

If a method first reads and then writes, the batch needs to be 
started before the read to ensure no changes can happen after 
the read but before the write. These changes make sure that
both reads and writes are covered in the same batch where 
required.

**This is needed on both master and 0.7.x**


csadilek:
@ederign assigning you again ;). these are the changes plus tests as discussed earlier today.

ederign:
Thanks!

Merged: [master] (https://github.com/uberfire/uberfire/commit/c861ef0e3a45a05b6cffa17fc8eea75a3cfe2122) [0.7.x](https://github.com/uberfire/uberfire/commit/79bcc46ea67a9f6aad148a8e974110c1c7c8d3e6)

We should for sure revisited that soon, at least with a execute around pattern to startbatch/endbatch.

PR #204, opened 2015-10-30
BZ1275325: WAS: indexing fails with ContextNotActiveException

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1275325

Implementation of EJB was flawed; it was neither an "no interface" bean, nor a "local interface" bean.

### DIFF
--- a/uberfire-commons/src/main/java/org/uberfire/commons/async/DisposableExecutor.java
+++ b/uberfire-commons/src/main/java/org/uberfire/commons/async/DisposableExecutor.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright 2015 JBoss Inc
+ *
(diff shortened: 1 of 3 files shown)


mswiderski:
just one comment we could add logger.debug into catch block here https://github.com/uberfire/uberfire/pull/204/files#diff-e9984cfa6dafef19bee2a91c5a06723cR67 to know the reason why it failed. As we might have some misconfiguration/library conflicts or so that will fail it on app servers while it should not be. Not a must to have but will simplify troubleshooting.

besides that looks good to me - +1

manstis:
@mswiderski Logging added.

mswiderski:
thank you

csadilek:
Thanks.

0.7.x: 71edcdf4462489c8bb676d745baf80f8d52b982e

PR #205, opened 2015-10-30
BZ1275325: WAS: indexing fails with ContextNotActiveException

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1275325

Implementation of EJB was flawed; it was neither an "no interface" bean, nor a "local interface" bean.

### DIFF
--- a/uberfire-commons/src/main/java/org/uberfire/commons/async/DisposableExecutor.java
+++ b/uberfire-commons/src/main/java/org/uberfire/commons/async/DisposableExecutor.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright 2015 JBoss Inc
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


psiroky:
Will this work on Tomcat, where there is no EJB implementation?

manstis:
@psiroky Hi, I've not changed the underlying mechanism to create an Asynchronous Executor. We try to lookup a EJB Asynchronous Bean [here](https://github.com/manstis/uberfire/blob/BZ1275325/uberfire-commons/src/main/java/org/uberfire/commons/async/SimpleAsyncExecutorService.java#L65) if it fails - as it would on Tomcat - we fallback to our own thread pool [here](https://github.com/manstis/uberfire/blob/BZ1275325/uberfire-commons/src/main/java/org/uberfire/commons/async/SimpleAsyncExecutorService.java#L71).

psiroky:
@manstis thanks for the explanation. I just saw "EJB" and immediately though about Tomcat....

csadilek:
Thanks.

master: 064be4dcd4cb322341d160e01fb9f458d59c4175

csadilek:
@jhrcek this is another good example where only an integration test would help guard against regressions on WAS. We have usage against the new interface in tests in kie-wb-common, but only an integration test can catch real problems here. Anyways, something for our list of ToDos.

PR #206, opened 2015-11-02
UF-243: Configurable locking strategy for workbench editors

csadilek:
It's now possible to use the lockingStrategy attribute
on @WorkbenchEditor to configure optimistic and pessimistic
locking on a per-editor basis. This allows developers to
turn off pessimistic locking and provide their own
locking solution.

### DIFF
--- a/uberfire-client-api/src/main/java/org/uberfire/client/annotations/WorkbenchEditor.java
+++ b/uberfire-client-api/src/main/java/org/uberfire/client/annotations/WorkbenchEditor.java
@@ -92,4 +92,18 @@ public @interface WorkbenchEditor {
      * is the trigger to create a new panel, if panel already exists this information is ignored.
      */
     int preferredWidth() default -1;
+    
+    /**
+     * Defines how and if locks are acquired when using this editor. By default, a pessimistic locking 
+     * strategy is used, allowing edits by only one user at a time.
+     */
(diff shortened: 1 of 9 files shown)


csadilek:
@manstis would you do me the favour and review this? We don't have a requirement for this but it seems like a good idea for UberFire community users who might not want the default pessimistic locking functionality we use in our workbenches. 

Nothing needs to change in the workbenches. We can still decide to remove the old optimistic locking code in the KIE editors as this is independent of this feature in UF.

manstis:
+1 looks good to me.. ideally I'd like to take it a stage further and have default "out of the box" support for either Optimistic or Pessimistic - UF now supports an OOTB implementation for Pessimistic locking; and we have support for an implementation of Optimistic locking in uberfire with the ObservablePath and related code. We probably need to consolidate APIs and implementations; with the default implementations being OOTB but replacable by Users. WDYT?

csadilek:
Thanks, @manstis! Yes, I agree. Let's refactor/move the optimistic locking functionality we have for KIE editors to UF so that community developers can more easily reuse it. We need to think about how an API for this would best look like.

csadilek:
Letting this sit for a few hours before I merge in case anyone else wants to review as well ;).

csadilek:
master: 9e48b048b7350a9155065ab2a978056da759a351

PR #208, opened 2015-11-09
UF-214 - Replaced listbar display style from table to flex

cristianonicolai:
Using display flex allows elements to be correctly sized inside the listbar toolbar header


csadilek:
+1

csadilek:
Thanks!

master: 134ead42feae96c9796c2ea6c8d22eacee2f9df3

PR #212, opened 2015-11-10
 Included selection when adding menus on WorkbenchMenuBar

cristianonicolai:
This fix is needed because in some cases, perspective change event is fired before menus is added.


csadilek:
+1

csadilek:
Thanks!

master: 99205e805b4baa1b3f972cdc646d836bc8afb449

PR #215, opened 2015-11-10
BZ1214245: Roles added to Organizational Unit and Repository are ignored by Business Central

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1214245

The issue was that a "silent" NPE was being thrown when diffs were calculated for files have do not have a parent commit (i.e. the file has not been changed). I am working on tests (to follow)... any help appreciated!

### DIFF
--- a/uberfire-nio2-backport/uberfire-nio2-impls/uberfire-nio2-jgit/src/main/java/org/uberfire/java/nio/fs/jgit/JGitFileSystemProvider.java
+++ b/uberfire-nio2-backport/uberfire-nio2-impls/uberfire-nio2-jgit/src/main/java/org/uberfire/java/nio/fs/jgit/JGitFileSystemProvider.java
@@ -56,6 +56,7 @@ import org.eclipse.jgit.lib.ObjectId;
 import org.eclipse.jgit.lib.Ref;
 import org.eclipse.jgit.lib.Repository;
 import org.eclipse.jgit.revwalk.RevCommit;
+import org.eclipse.jgit.revwalk.RevTree;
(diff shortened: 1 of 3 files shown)


manstis:
@porcelli Please review.. I am trying to write a test, but setting up SSH in a test is proving problematic.

manstis:
@porcelli [This](https://github.com/manstis/uberfire/blob/BZ1214256/uberfire-nio2-backport/uberfire-nio2-impls/uberfire-nio2-jgit/src/test/java/org/uberfire/java/nio/fs/jgit/JGitFileSystemProviderSSHTest.java) is the test I have.. but it fails with a ```org.eclipse.jgit.api.errors.InvalidRemoteException: Invalid remote: origin``` exception.. I have no idea what I'm doing wrong :(

manstis:
@porcelli Unit Test now added :)

csadilek:
@manstis looks good, but the test passes with and without the fix. So, it's not covering your changes.

manstis:
I debugged the test and the commit pushed to origin didn't have a parent so
I'd have thought it'd have thrown a RTE... but if you're saying otherwise I
guess I need revisit :-(

csadilek:
OK we have to make a compromise here. The PR clearly fixes this issue and adds test coverage. We can't currently cover all against regressions cases and created a JIRA to do so on master. The refactoring do make this possible is too invasive for 0.7.x: https://issues.jboss.org/browse/UF-247

csadilek:
Thanks!

master: 08c9277362fefe1ff6af694bcfd3e7f0c6cbc6d6

PR #227, opened 2015-11-24
[URGENT] Fix for cluster issue BZ-1283953

porcelli:
BZ-1283953: Cluster failing to cleanup the global lock properly. After investigating, it was noticed that some FileSystem ("system") were proxyed, and the CDI proxy doesn't include other possible interfaces other than that been injected... so when tested instance of FileSystemId it always failed.
So.. in order to avoid such issue, the solution is to avoid use directly the file system and resolve it using the fs.getRootDirs().interator().next().getFileSystem(); with this we assure that we have access the real FileSystem instance.

Additional PR:
https://github.com/droolsjbpm/guvnor/pull/199

### DIFF
--- a/uberfire-io/src/test/java/org/uberfire/io/impl/cluster/IOServiceClusterImplTest.java
+++ b/uberfire-io/src/test/java/org/uberfire/io/impl/cluster/IOServiceClusterImplTest.java
@@ -1,0 +1,128 @@
+/*
+ * Copyright 2015 JBoss, by Red Hat, Inc
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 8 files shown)


uberfire-io/src/test/java/org/uberfire/io/impl/cluster/IOServiceClusterImplTest.java
```diff
@@ -0,0 +1,128 @@
...
+
+            ioServiceCluster.startBatch( new FileSystem[]{ mockedFS }, mock( Option.class ) );
+
+            assertEquals( 1, ioServiceCluster.batchFileSystems.size() );
+
+            assertTrue( ioServiceCluster.batchFileSystems.contains( ( (FileSystemId) mockedFSId ).id() ) );
+
+            ioServiceCluster.endBatch();
+
+            verify( serviceLockable, times( 2 ) ).endBatch();
```
manstis: Is this 2 because your mocks are not reset (and this is the second test)? In the third test you check ```endBatch``` was called 3 times...

manstis:
+1 to merge.

csadilek:
Thanks!

0.7.x: 4dbd922dc942b0801d0f1bae0ab19fbae009a595

master: 0d9af94fd979a9eaeadc7a9c11e036d83acafa79 (did that by mistake, but since it belongs there anyway I will leave it. We don't need another PR for master here but for Guvnor).

PR #231, opened 2015-11-25
[Urgent] BZ-1266138: managed beans can't be used during container lifecycle on WAS.

porcelli:
BZ-1266138: managed beans can't be used during container lifecycle on WAS. Code is not nice, but is the only way that I managed to make it work on WAS without changing all architecture.

Important note: this "code" is the only way to make it work on WAS that i found.


csadilek:
+1

manstis:
Yuck... and no tests, but given that'd require WAS integration tests (which isn't going to happen any time soon) I bow to your investigation and +1 to merge.

porcelli:
@manstis tests here is to test would be something like: object instanceof Object :D

csadilek:
Thanks.

0.7.x: fb19e38b19186003ea482f6979e3284650358af9

csadilek:
Testing for WAS, WLS is an ongoing topic. Integration-test are badly needed. We might be able to test that we put the systemFs into the PriorityDisposables map in a unit test....but we need to investigate that for master (out of time and this needs to get in).

manstis:
"Write once, run (almost) anywhere"

PR #237, opened 2015-12-01
Upgrade mocking deps to fix test on JDK8

psiroky:
* fixes failing test DisposableShutdownServiceTest
 * PanelManagerImpl moved from field injection
   to constructor injection so that @InjectMocks
   could be removed (was causing issues)
* few minor fixes related to exclusion of old javassist dep

There are still three tests failing on JDK8 (`TabPanelWithDropdownsTest`), but they were failing even before these changes and I am currently not sure how to fix them. We need to take a loot at them later.

I tried local build with JDK7 and everything works as expected (no failures).

### DIFF
--- a/uberfire-api/pom.xml
+++ b/uberfire-api/pom.xml
@@ -47,6 +47,12 @@
       <groupId>org.jboss.errai</groupId>
       <artifactId>errai-security-server</artifactId>
     </dependency>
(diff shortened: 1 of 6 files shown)


csadilek:
+1

csadilek:
Thanks.

master: 60e30823d702e685e1525bb8ee848b1603da703d

PR #238, opened 2015-12-02
CallerProxy - Return primitive default value in case of exception

cristianonicolai:
This is to avoid a NullPointerException on Proxy instance when casting the returned null value into the primitive return type.


manstis:
@ederign Could you please review?

ederign:
+1

csadilek:
master: 4800a5b0af082bbbe6c416d07d1ba8b2af7e061a

PR #241, opened 2015-12-04

porcelli:


csadilek:
+1, only concern are the Thread.sleep statements in the test which don't seem to be needed.

manstis:
+1 to merge.

csadilek:
Thanks.

master: 8dd635d2308d9569adccaab13dad36900a00c5b8

PR #242, opened 2015-12-04

porcelli:


porcelli:
https://bugzilla.redhat.com/show_bug.cgi?id=1285845

uberfire-nio2-backport/uberfire-nio2-impls/uberfire-nio2-jgit/src/main/java/org/uberfire/java/nio/fs/jgit/JGitFileSystemProvider.java
```diff
@@ -212,9 +214,8 @@
     private final Map<JGitFileSystem, Map<String, NotificationModel>> oldHeadsOfPendingDiffs = new HashMap<JGitFileSystem, Map<String, NotificationModel>>();
 
     private Daemon daemonService = null;
+
     private GitSSHService gitSSHService = null;
-    private FileSystemAuthenticator authenticator;
```
manstis: Why are ```FileSystemAuthenticator``` and ```FileSystemAuthorizer``` removed too (these having nothing to do with SSH timeouts..) Did they server no purpose before (SSH authentication is not at the FS level...)?

porcelli: removed as they weren't used. the only use of it was for ssh.. that are set with a setter... so we were just holding a field here for nothing

manstis:
+1 to merge.

csadilek:
0.7.x: ad73d43f10123adcb6c6c8f3d4e5fd387e5e818b

PR #243, opened 2015-12-04
Fix TabPanelWithDropdownsTest for JDK8 builds

csadilek:
- There are multiple dependent scoped beans satisfying
TabPanelWithDropdowns. Which one is returned by Errai's
beanmanager is not deterministic and dependent on the
order of type discovery.


csadilek:
@psiroky test fix, as discussed. please review.

psiroky:
+1 to merge. Local build with JDK8 passed.

Just out of curiosity, why were those `com.google` imports moved after the `org.` ones?  Is it on purpose or just misconfiguration?

csadilek:
@psiroky that's how Eclipse formatted it. I am using the official formatter: https://github.com/uberfire/uberfire/blob/master/drools_jbpm_eclipse_formatter_config.xml

psiroky:
Hmm, I would expect it to order the imports alphabetically. Anyway, this is just minor detail.

csadilek:
master: 508181b7bfcf5eb0c472a023989eff88f0c1167d

PR #245, opened 2015-12-04

porcelli:


porcelli:
https://bugzilla.redhat.com/show_bug.cgi?id=1283837

manstis:
What no tests!!?

manstis:
@csadilek @porcelli If we agree, since 0.5.x is as good as dead, we don't need tests I can +1..

porcelli:
for sure this is not a dead branch - as we have product on this. but the existing tests for this feature on master (and 0.7.x) would demand breaking existing code or create new ones from scratch.
so... I considered less harm keep tests on master and 0.7.x (the most active branches) and backport just the minimal - specially for a simple change like this ;)

manstis:
+1 to merge.

csadilek:
0.5.x: eb77668c6df1f12b09bcd83a3d82715931f9241c

PR #247, opened 2015-12-07
BZ-1287511: Fixes

porcelli:
The fix is splitted in 2 commits on purpose.

1st commit avoids empty amends (basically avoid recreate history for batch on branches that didn't have a commit)
2nd commit avoids block the cleanup of historic notification if excaption happens during notification (a good case is the missing tree reported by BZ)

important note: was barely impossible to reproduce the Missing tree error. A quicl google search shows that people that has faced such issue had hard time (to not say couldn't) reproduce it.


csadilek:
+1

csadilek:
Thanks! ;)

0.7.x 2897ef81c7b8549ae8689ccf6be8109a46c9baa0 de5fa01c818262f7e07b53ebe267f96ff867f1ce

PR #248, opened 2015-12-07
BZ-1287511: Fixes for master

porcelli:
The fix is splitted in 2 commits on purpose.

1st commit avoids empty amends (basically avoid recreate history for batch on branches that didn't have a commit)
2nd commit avoids block the cleanup of historic notification if excaption happens during notification (a good case is the missing tree reported by BZ)

important note: was barely impossible to reproduce the Missing tree error. A quicl google search shows that people that has faced such issue had hard time (to not say couldn't) reproduce it.


csadilek:
master: 62517b8be9c724adebb82cd85497015e809ce768 e2429f2e8eec736707b8956795bd9ce41083623b

PR #252, opened 2015-12-09
Add animal-sniffer to catch the usage of JAVA 7+ API

psiroky:
* added into "full" profile as it slows down the build a bit

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -101,6 +101,37 @@
           </executions>
         </plugin>
         <plugin>
+          <groupId>org.codehaus.mojo</groupId>
+          <artifactId>animal-sniffer-maven-plugin</artifactId>
+          <version>1.13</version>
+          <dependencies>
+            <!-- Upgrade ASM and support Java 8 bytecode -->
+            <!-- This is a workaround as it is already fixed in version 1.14.
+                 However, that one fails with NPE for even simple cases, so it can't be used -->
+            <dependency>
+              <groupId>org.ow2.asm</groupId>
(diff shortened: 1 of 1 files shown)


csadilek:
We plan to upgrade to GWT 2.8 for the next/next release. Then we want to be using Java 8 syntax and APIs on the client (compiled to JS). On the server we can still keep compatibility with Java 6 or whatever we need at that point. Question is, can the plugin be configured to verify based on packages or should we add it in for now and remove it later?

psiroky:
I believe it should be possible to exclude/ignore the client classes from the check. Plugin has `ignores` configuration option were we can specify "Class names to ignore signatures for (wildcards accepted)" (http://www.mojohaus.org/animal-sniffer/animal-sniffer-maven-plugin/check-mojo.html#ignores).

csadilek:
OK we try and configure once we need it, +1.

psiroky:
Yes. In case it would prevent usage of Java 8 APIs in client classes, we can remove the configuration.

csadilek:
master: c04905f9b111cc679f6bafb793bd2a61faf4517f

manstis:
Uberfire and uberfire-extensions also lack the transient dependency enforcer check...might be worth adding for consistency with drools/jbpm. WDYT?

psiroky:
It is there, but for now reporting only (https://github.com/uberfire/uberfire/blob/master/pom.xml#L64). We need to clean-up the modules to not rely on transitive dependencies. After the we should be able to enable it completely (as it would fail the build if violation found).

PR #255, opened 2015-12-10
Allow modal footer to be customized

cristianonicolai:
Only add default footer in case a container using modal-footer class isn't provided


manstis:
How, where, why does this compare to BaseModal IIRC we had floating around somewhere and used as the Base for all (most) popup in (at least) KIE Workbench etc?

cristianonicolai:
@manstis true, most cases use BaseModal AFAIK. Problem is that I found some cases ( ex: ServerRegistryEndpointPresenter, ProcessInstanceSignalPresenter ) where @WorkbenchPopup still in use and this screens have duplicated action buttons. The ones defined in the pop-up plus the default OK one. This is the simplest fix I could think of without a major refactor to use BaseModal.

manstis:
@cristianonicolai OK, to be honest I cannot see why the *old* code did not work correctly; but that said your new code does much the same thing in a round about way and, if it fixes the issue, then +1 to merge!

csadilek:
master: 31d658d4ff1c1ed35285687095f7d8922987456b

PR #257, opened 2015-12-14
Remove forgotten uberfire-workbench-client-views-bs2 references

psiroky:
This is quick&dirty fix as it basically just makes sure the stuff compiles. I am not exactly sure what will happen to the application after removing the -bs2 dependency. It seems the proper fix would be to migrate all the stuff BS2 stuff to BS3/Patternfly, but looks very complicated for me...

@mbiarnes this should remove all the references to -views-bs2 module + fix the compilation error you saw (just added the gwt-bootsrap dependency).

### DIFF
--- a/uberfire-bom/pom.xml
+++ b/uberfire-bom/pom.xml
@@ -164,17 +164,6 @@
       </dependency>
       <dependency>
         <groupId>org.uberfire</groupId>
-        <artifactId>uberfire-workbench-client-views-bs2</artifactId>
-        <version>${project.version}</version>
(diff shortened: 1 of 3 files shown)


csadilek:
@psiroky @mbiarnes I am working on a fix in parallel (just had some Eclipse issues to deal with :)). I will see if I can upgrade the showcase to BS3 quickly. If not, I will merge this one.

csadilek:
@psiroky ok sending separate PR and closing this one. Thanks.

PR #258, opened 2015-12-14
Adding back version for m2e lifecylce-mapping plugin

csadilek:
- Removing it causes lots of NPEs (in maven paring logic) and
prevented Eclipse from building the project correctly.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -330,6 +330,7 @@
         <plugin>
           <groupId>org.eclipse.m2e</groupId>
           <artifactId>lifecycle-mapping</artifactId>
+          <version>1.0.0</version>
           <configuration>
             <lifecycleMappingMetadata>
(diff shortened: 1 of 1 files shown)


psiroky:
Looks like it is a bug directly in Eclipse m2e: https://bugs.eclipse.org/bugs/show_bug.cgi?id=371618

I am OK with adding the version back, but we should add a comment explaining why the version is there (as it is already defined in parent). Otherwise it may very well happen that my next "clean-up" may remove it again...

psiroky:
Added the comment and merged into master. Sorry about this -- I had no idea Eclipse would not pick the version correctly from parent.

csadilek:
@psiroky thanks and no worries, it's an annoying bug in Eclipse (m2e) that's been around for too long.

PR #260, opened 2015-12-14
Fix test compile error in jdt/eclipse

csadilek:
.when() was ambiguous (imported statically from
both Mockito and PowerMockito)


csadilek:
@psiroky when you have the time, please ;).

PR #261, opened 2015-12-24
[UF-261] jbpm console's RepositoryEditor view does not reflect the co…

abhijithumbe:
Proposal fix for UF-261 . Please review.


porcelli:
Sorry, this is not a fix. Actually this is not even a bug, you should use port and hostport. Would be good to have a PR from you that would apply for ports the same logic used by host and hostname.
Anyway thank you @abhijithumbe - hope to see a new PR from you soon!

PR #266, opened 2016-01-05
GUVNOR-2328: Asset Editor height doesn't count with the added top tabs

manstis:
See https://issues.jboss.org/browse/GUVNOR-2328

The cause is that the "tab content" region's size is set to 100% which is the size of the ```DIV``` containing both the "tab header" and "tab content" regions. This fix ensures the "tab content" size is 100% less the height of the "tab header".


manstis:
See https://github.com/droolsjbpm/drools-wb/pull/106 too.

manstis:
@porcelli @ederign @cristianonicolai @csadilek Does anybody know of any other UF "containers" that could possibly suffer the same issue?

csadilek:
+1

csadilek:
master: 7219b51b1eab277aba2f1be3b03fb84816b9a967

PR #267, opened 2016-01-06
UF-258: Plugins can be copied or renamed to non-editable plugin names

paulovmr:


paulovmr:
PR revised by Hangouts with @ederign. I will do some modifications on it.

paulovmr:
The creation of the ParameterizedReturnCommand interface was unnecessary.

PR #270, opened 2016-01-08
Add message to thrown NoSuchFileException.

zkrejcov:
Unless the exception message was left empty for some good reason, this seems like a good idea.. ;-)


csadilek:
Thanks!

master: 8454327ce5cf5194c0927c1df9f5196e4c4450ae

PR #271, opened 2016-01-11
UF-264 : TempFiles class for test utils

Rikkola:
Something we quite often need. Simplifies the setup. 
We could place this into a drools/jbpm test util module, but we do not have one. This functionality is usually needed when making real paths/uberfire projects for tests. I see myself committing utility classes for an Uberfire filesystem set up later, they will depend on this class.

### DIFF
--- a/uberfire-testing-utils/src/test/java/org/uberfire/files/TempFilesTest.java
+++ b/uberfire-testing-utils/src/test/java/org/uberfire/files/TempFilesTest.java
@@ -1,0 +1,44 @@
+package org.uberfire.files;
+
+import java.io.File;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+public class TempFilesTest {
+
+    private TempFiles tempFiles;
+
+    @Before
+    public void setUp() throws Exception {
+        tempFiles = new TempFiles();
+    }
+
+    @Test
+    public void testDirectory() throws Exception {
+        final File tempDirectory = tempFiles.createTempDirectory( "mydir" );
+
+        assertTrue( tempDirectory.exists() );
+
+        tempFiles.deleteFiles();
+
+        assertFalse( tempDirectory.exists() );
+    }
+
+    @Test
+    public void testFiles() throws Exception {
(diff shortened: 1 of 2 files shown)


uberfire-testing-utils/src/test/java/org/uberfire/files/TempFilesTest.java
```diff
@@ -0,0 +1,44 @@
...
+        final File pomFile = tempFiles.createTempFile( "mydir/pom.xml" );
+        final File javaClass = tempFiles.createTempFile( "mydir/org/test/Person.java" );
+
+        assertTrue( pomFile.exists() );
+        assertTrue( javaClass.exists() );
+
+        tempFiles.deleteFiles();
+
+        assertFalse( pomFile.exists() );
+        assertFalse( javaClass.exists() );
```
manstis: Is it worth checking ```mydir``` doesn't exist too? Otherwise if a subsequent test tried to create temp file ```mydir``` would ```TempFiles``` throw an exception (I did not see that it quietly failed if a folder already existed).

manstis:
Other than my comment above; +1 to merge.

uberfire-testing-utils/src/test/java/org/uberfire/files/TempFilesTest.java
Rikkola: It probably is a good idea, better to make this as good as possible.

mrietveld:
@Rikkola We know have this: https://github.com/droolsjbpm/drools/tree/master/kie-test-util

..although I don't know if that's the type of module you meant?

Rikkola:
Thanks @mrietveld had forgotten about that.
I could put this class there, but since I need this to make a helper class for testing Uberfire FileSystem it adds the dependency for Uberfire into drools. Based on this Uberfire is a better location.

Rikkola:
Going to place this into guvnor for now.
1. I saw that Uberfire already has some sort of test util class for managing the file system.
2. This is mainly just used by other test util classes that make more sense in guvnor/kie-wb-common

PR #273, opened 2016-01-12
Fix duplicated classes + add enforcer rule to detect these automatically

psiroky:
* fix to make sure there is always only one (with excpetions) class
   with the particular FQN on classpath

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -76,7 +76,15 @@
     <pluginManagement>
       <plugins>
         <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-enforcer-plugin</artifactId>
+          <dependencies>
+            <dependency>
+              <groupId>org.codehaus.mojo</groupId>
+              <artifactId>extra-enforcer-rules</artifactId>
+              <version>1.0-beta-4-jbossorg-1</version>
+            </dependency>
+          </dependencies>
           <executions>
             <execution>
               <id>ban-unwanted-logging-deps</id>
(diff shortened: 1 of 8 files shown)


psiroky:
I removed some Errai deps excludes as they were recently done directly in Erra and are no longer needed in UF.

csadilek:
+1

pom.xml
```diff
@@ -97,6 +105,60 @@
...
+                           See https://github.com/errai/errai/blob/master/errai-javax-enterprise/README -->
+                      <ignoreClass>javax.interceptor.InvocationContext</ignoreClass>
+                      <ignoreClass>javax.inject.Provider</ignoreClass>
+                      <ignoreClass>javax.inject.Named</ignoreClass>
+                      <ignoreClass>javax.inject.Scope</ignoreClass>
+                      <ignoreClass>javax.inject.Qualifier</ignoreClass>
+                      <ignoreClass>javax.enterprise.*</ignoreClass>
+                      <!-- Classes from gwt-user duplicated in errai-uibinder -->
+                      <ignoreClass>com.google.gwt.uibinder.rebind.UiBinderWriter</ignoreClass>
+                      <ignoreClass>com.google.gwt.uibinder.rebind.UiBinderGenerator</ignoreClass>
```
csadilek: @psiroky the goods news is we deleted that module in Errai 4. So, as soon as we upgrade we can delete those.

csadilek:
master: 885dd93b6c6d94d962cab9465683df14662bb517

PR #284, opened 2016-02-02
UF-274 Plugin Explorer doesn't load: dynamic only beans should be @al…

ederign:
…ternative to never be avaliable for injection or resolution


csadilek:
+1

csadilek:
master: 91d6244e7e67ce09eb52d5bf833ecfceeb21d1c1

PR #290, opened 2016-02-04
Scrollbars always displayed on latest version of Chrome (48)

ederign:
Scrollbars are always displayed on Google Chrome 48. After further investigation, we figure out that this only happens on OSX + Chrome 48 + Big Screens. See the magic: http://cl.ly/2x0o0P1h3w3X

Uberfire's uses a Header Panel, that creates a container to wrap the header widget. This container has  a ResizeLayoutPanel inside that has "overflow: scroll" defined and chromes 48 thinks that it's a good idea to show them in biggers screens (only on osx). (╯°□°）╯︵ ┻━┻

https://github.com/google-web-toolkit/gwt.svn/blob/master/user/src/com/google/gwt/user/client/ui/HeaderPanel.java#L71
https://github.com/gwtproject/gwt/blob/master/user/src/com/google/gwt/user/client/ui/ResizeLayoutPanel.java#L163

See this gist: https://gist.github.com/ederign/79a10546fa60277ac587#file-gistfile1-txt-L25

In order to fix that, I've to add a id on this header panel in order to access inner div with a css selector.

@cristianonicolai could you please take a look on it and see if you figure out a better solution?

### DIFF
--- a/uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/resources/org/uberfire/client/views/static/uberfire-patternfly.css
+++ b/uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/resources/org/uberfire/client/views/static/uberfire-patternfly.css
@@ -294,3 +294,8 @@
 .form-horizontal .uf-form-label {
     text-align: right;
 }
+
+.uf-workbench-layout > div:first-child > div:first-child,
+.uf-workbench-layout > div:first-child > div:nth-child(2) {
+    overflow: hidden !important;
+}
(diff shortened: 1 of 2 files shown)


ederign:
And thanks @csadilek and @paulovmr  for the help.

cristianonicolai:
@ederign overall seems fine, I only find one problem. If you apply overflow:hidden to all child divs, the menu bar dropdown menu ends-up hidden. So only the first two divs child of `#headerContainer` should be changed. I made a new commit using slightly different approach, have a look if works for you: https://github.com/cristianonicolai/uberfire/commit/2bd700ba5f3ad580d66c879fa6d152cdb5e438eb

ederign:
@cristianonicolai thanks for the review and to pointing out a better solution! Still works!

PR updated.

csadilek:
+1, great teamwork to tame this scrollbar ;).

csadilek:
master: 8e8e586a0540c5b749f20492c944d54575aef94f

@ederign can you send this to 6.4.x as well, thanks.

PR #294, opened 2016-02-08
Remove no longer needed javassist exclusions/inclusions

psiroky:
* javassist:javassist is now properly excluded directly in the
   Errai, and replaced by org.javassist:javassist

 * we still have the enforcer rule in place, in case it would
   somehow got back. The build would fail in that case.


manstis:
+1 to merge.

csadilek:
master: 8d7ce8b52bbc060387dc3600a82e6ad2c06033de

PR #295, opened 2016-02-16
upgraded jboss-ip-bom

mbiarnes:


manstis:
+1 to merge.

psiroky:
The build failed because the ip-bom-6.0.2.Final is not yet synced into Maven central (can take up to one day after releasing). So we either need to wait of add the repository.jboss.org into the pom.xml (like it is already done in other repos).

psiroky:
I've re-triggered the build. Once it succeeds I will merge.

psiroky:
Merged into 0.8.x

PR #296, opened 2016-02-17
[0.8.x] UF-201: Notifications interfere with top menu navigation

paulovmr:


uberfire-workbench/uberfire-workbench-client/src/main/resources/org/uberfire/client/resources/css/workbench.css
```diff
@@ -18,7 +18,7 @@
     border: 0;
     margin: 0;
     padding: 0;
-    z-index: 100000;
+    z-index: 990;
```
manstis: Are you confident ```990``` is higher than the ```z-order``` used by ```BaseModal``` otherwise the notifications will appear behind modal dialogs?

paulovmr: No, I am not. Changing it now. Thanks.

paulovmr: @manstis I changed it. Now the menu stays above the notification.

manstis:
If I understand correctly the first (top most) notification will be positioned below the top most menu bar and dropdown menus in that bar will appear over notifications. Seems reasonable.... however the issue then moves to secondary menu bars... but that I guess is the topic of a different (future) BZ ;-)

manstis:
+1 to merge.

PR #298, opened 2016-02-24
BZ1299619: Business Central Repo Clone Fails To Clone SCP Style SSH URLs

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1299619

This PR upgrades ```uri.min.js``` to support ```git@github.com:<user>/<repository>.git``` style URL validation and support for SSH Passphrases when cloning. I cannot see any feasible way to write a test for this - although I've confirmed operation with ```kie-drools-wb``` (both cloning a ```github``` Repository with ```SCP``` URL and cloning a Workbench repository with ```ssh://user@localhost:8001/uf-playground```). 

Steps to test:-

1) Create a new Repository at github.com
2) Create an SSH key; see [here](https://help.github.com/articles/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent/) (a)
3) Add your SSH key to github; see [here](https://help.github.com/articles/adding-a-new-ssh-key-to-your-github-account/)
4) Run ```kie-drools-wb``` (or which ever you fancy)
5) Clone a Repository using the ```git@github.com:<user>/<repositorty>.git``` URL

(a) *IF* you entered a passphrase when generating your public key you'll need to specify the passphrase with System Property ```org.uberfire.nio.git.ssh.passphrase```


manstis:
@porcelli You're probably best to review...

Rikkola:
Looks like the uri.min.js is a generated file. Just wondering we need to commit it in or if the build recompiles it each time?

Rikkola:
+1 
uri.min.js can be removed later. It is already in. If it even needs to be removed.

manstis:
@Rikkola it comes from [here](https://github.com/garycourt/uri-js/tree/master/dist). It was included in the previous version; so I simply updated... At the moment it's considered a ```GWT``` ```TextResource``` and injected into the host page. If we removed from the Uberfire module we'd need to add a ```<script>``` element in each webapp's host page. This approach, whilst bundling the source in our source, means it's self-contained and "black boxed".

manstis:
@psiroky @mswiderski Could you please merge when you get a chance... (I need to submit additional PRs for 6.4.x and 6.3.x once merged into master).

mswiderski:
@manstis I don't have access to push to uf repo :(

porcelli:
isn't passphrase per certificate? If yes, wouldn't this limit users to operate ssh (scp style) with a single remote entity?

manstis:
@porcelli Damn you.. don't you have some Execution Server tests to write? ;) 

You are of course correct, but here's my take on it: The public certificate is to set up a trust relationship between a source (the server running the Workbench) and other end-points (git servers/github repositories). So a single "server" public certificate is required that could be added to different end-points and hence a single pass-phrase.

If the Workbench ran under different user accounts, depending who's logged in, we would need to provide the means to interactively request a pass-phrase as the SSH key resolution would be different per-user.. but we don't work like this. The Workbench makes clone requests as a single privileged user (i.e. "the system"). Or, if "the system" wanted to identify itself differently for different end-points we'd need to support the same... but this, IMO, would be a questionable requirement.

porcelli:
@manstis I understand. Althought I don't think this is the ideal solution, it might be good enought for this particular use case (that originated BZ).
 I just would make it clear on docs.
(the real solution for that would be way more complex, basically enabling UF to m`manage` certificates and passphrases, what for sure would need a lot of careful as it implies in possible security flaws).

PR #309, opened 2016-03-07
BZ1310893: BxMS6.2.0 Full Text search returns only 10 results even if it hits more than 10

manstis:
See https://bugzilla.redhat.com/show_bug.cgi?id=1310893

```IOSearchService``` has been moved to uberfire-extensions (as changes for the BZ need access to ```KObject``` class).


manstis:
Part of an ensemble:

https://github.com/uberfire/uberfire/pull/309
https://github.com/uberfire/uberfire-extensions/pull/226
https://github.com/droolsjbpm/guvnor/pull/275
https://github.com/droolsjbpm/kie-wb-common/pull/288
https://github.com/droolsjbpm/drools-wb/pull/137
https://github.com/droolsjbpm/kie-wb-distributions/pull/218
https://github.com/droolsjbpm/optaplanner-wb/pull/43
https://github.com/droolsjbpm/jbpm-console-ng/pull/356
https://github.com/droolsjbpm/jbpm-designer/pull/177

manstis:
I will need to cherry-pick this to 6.4.x and 6.3.x (for https://bugzilla.redhat.com/show_bug.cgi?id=1303534)

csadilek:
+1 (whole series). @manstis Not related to this PR, but I think we should add an application scoped producer to kie-wb-common that provides this default search service instance (and others) for all web apps instead of duplicating all this in the various (ApplicationScopedProducer|ApplicationScopedProvider)s. We can mark it as ```@Alternative``` so that customization in the various webapps is still possible. Then a webapp only needs to provide an app scoped producer if it needs a non-default configuration. This would make changes like this a lot easier in the future. Wdyt?

csadilek:
master: 88e5867857b527b7be870310eabde422ca17231c

PR #315, opened 2016-03-09
GUVNOR-2362: Project Explorer freezes after uploading an item with an extension-free name

manstis:
See https://issues.jboss.org/browse/GUVNOR-2362

This PR removes the server-side file name mashing... it is now performed client-side.

The issue was that the "corrected" path is not sent back to the client; and when the editor loaded the path used to launch the editor is incorrect and hence loading fails...

### DIFF
--- a/uberfire-server/src/main/java/org/uberfire/server/FileUploadServlet.java
+++ b/uberfire-server/src/main/java/org/uberfire/server/FileUploadServlet.java
@@ -58,28 +58,17 @@ public class FileUploadServlet
 
                 final FileItem fileItem = getFileItem( request );
 
-                finalizeResponse(response, fileItem, uri);
+                finalizeResponse( response, fileItem, uri );
 
             } else if ( request.getParameter( PARAM_FOLDER ) != null ) {
 
-                //See https://bugzilla.redhat.com/show_bug.cgi?id=1091204
-                //If the User-provided file name has an extension use that; otherwise use the same extension as the original (OS FileSystem) extension
-                String targetFileName;
-                final FileItem fileItem = getFileItem( request );
-                final String originalFileName = fileItem.getName();
-                final String providedFileName = request.getParameter( PARAM_FILENAME );
-                if ( providedFileName.contains( "." ) ) {
-                    targetFileName = providedFileName;
-                } else {
-                    targetFileName = providedFileName + getExtension( originalFileName );
-                }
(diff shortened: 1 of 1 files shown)


manstis:
Part of an ensemble:-

https://github.com/uberfire/uberfire/pull/315
https://github.com/uberfire/uberfire-extensions/pull/236
https://github.com/droolsjbpm/kie-wb-common/pull/299

csadilek:
@manstis test failure is related to this commit, (un)fortunately ;).

manstis:
Jenkins please retest this.

manstis:
@csadilek Heh.. I don't see how changing the source code for a ```Servlet``` affects the "usable free space" on the servers FileSystem ;) -- tests run OK locally.. the failing line is effectively " File.listRoots()[ 0 ].getUsableSpace()== File.listRoots()[ 0 ].getUsableSpace();" which points to a concurrency issue between two reads of the OS's FileSystem.. I've asked Jenkins to retest.. let's see if he's in a good mood today.

manstis:
@csadilek The concurrency Gods smile on me today ;)

csadilek:
@manstis hm. I was getting the failure after cherry-picking the commit but not without it. Didn't look at the  test in detail though. I will put that on my Todo list. Seems it's another test we can't rely on.

csadilek:
master: 61922b92a984e66a8d1f39a49d264cf1dea36d64

manstis:
@csadilek Our Jenkins CI environment runs builds (possibly only tests) in parallel.. sometimes we can control things (e.g. a lot of tests involving JGIT don't play nicely concurrently - as ```JGitFileSystemProvider``` automatically starts the Git daemon and ssh). These we can control. However this failing test was really weird, it reads the underlying OS's FileSystem "free space" natively and the same through Uberfire's VFS comparing the two.. The only way it could differ would be if the OS's FileSystem "free space" changed in between the two reads... and IDK if we can synchronize OS actions from Java (e.g. one POSIX thread could be running Java and hence the test and another POSIX thread doing some file writes).

csadilek:
@manstis I was reproducing the problem on my local machine without parallel builds. We don't need to add fancy synchronization there, I believe. Removing the assertion on usable FS space would be better then. It's on my Todo list but that's a looong list. Might get to it before I retire though ;).

PR #316, opened 2016-03-10
Remove outdated jacoco-gwt-maven-plugin

psiroky:
* it does not support Java 8. It can be added
   later back once that gets fixed

The other option is to fix the plugin (https://github.com/errai/jacoco-gwt-maven-plugin right away and upgrade it.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -227,12 +227,6 @@
         </plugin>
 
         <plugin>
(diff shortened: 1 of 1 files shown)


manstis:
+1 to merge.

csadilek:
master: 3e0d1f7e67c15749ab36e8d28866e63fb37a23b7

PR #318, opened 2016-03-11
GUVNOR-2332 - Allow context menu to be displayed without selecting a …

cristianonicolai:
…menu item


manstis:
+1 to merge.

cristianonicolai:
Do not merge this PR yet, I found an issue with perspectives not been started when I tried to get the context menus.

cristianonicolai:
ok, ready to go.

@manstis you might want to review it again, thanks.

manstis:
Any reason we cannot inject PlaceManager into WorkbenchMenuBarPresenter rather than use IOC'S static methods? It'll be easier to test if we remove use of IOC directly - especially as the latest Errai SNAPSHOTs change IOC making it unusable in tests (or code called by tests)... see @csadilek's PR on uberfire.

csadilek:
In this case an injected PlaceManager can be used it instead to avoid the programmatic lookup (but we can keep that separate). IOC is still testable using gwt-mockito or gwt-testcase, but no longer using plain unit tests as we have to reference a type with JSNI methods.

csadilek:
+1

csadilek:
0.8.x: eae1f4b6084dd9c92341c1deddc033a86735a0a4

PR #321, opened 2016-03-15
UF-280: Authz Management

dgutierr:
This is the first PR towards the full implementation of the following ticket:

https://issues.jboss.org/browse/UF-280

.. and it also contains this one:

https://issues.jboss.org/browse/UF-290

The PR contains the authorization management APIs, client/backend implementations and tests. In this very first milestone the following features are supported:

* A security policy can be defined at webapp level. For example: 

https://github.com/dgutierr/uberfire/blob/UF-280/uberfire-showcase/uberfire-webapp/src/main/resources/security-policy.properties

The security policy allows for declaring permissions over managed resources (currently only perspectives, screens, editors, splash screens, popups or menu entries). 

Every webapp can provide a custom policy (a file named _security-policy.properties_ available under the webapp's classpath). If no policy is available then the UF security module will behave **exactly** as before the new API was added, i:e. checking the user's roles against the roles attached to Resource instances.

* Permissions can be granted/denied either to a set of resources or over a single resource. For example:

  "Users with role analyst can view all perspectives except _Administration_"

        role.analyst.permission.perspective.view.*=true  
        role.analyst.permission.perspective.view.Administration=false

* The permission types can be easily extended to support in the future other workbench resources like repositories, projects or organizational units, just by adding new _PermissionType_ implementations.

* Ability to set the _Home_ perspective the user will be redirected after login.

           role.admin.home=AdminHomePerspective

Next steps to fully implement the UF-280 ticket are:

* Security policy editor to let users to modify the security policy right from the UI
* Security policy storage under the UF's system repo
* Support for filesystem/repository/path permissions 
* Documentation 

@porcelli Can you review & merge please?


manstis:
@dgutierr @porcelli Since (I assume) this is for 7.x is it best to dispense of the old ```@Role``` etc based authorization mechanism - rather than fallback to it as these PRs do? If it's acknowledged the existing mechanism is insufficiently flexible isn't 7.x the best time to replace it entirely?

porcelli:
@manstis maybe use @Role to collect data during compilation and generate the policy files

porcelli:
@dgutierr as I mentioned before, I have serious doubts how well the current single policy file will work for us. if we just consider the current state of workbench ecosystem we'll have hundres and potentially thausands lines in those policy files. And we'll also have to mantain those.
For 7 we have plans for Apps, and I'm sure apps will this ACL feature... so we'd need a mechanism to support runtime modularity.

And for our codebase I'd like to see, in order to avoid us having to define policy files manually, collect all those data from possible annotations.

manstis:
@porcelli sure, it'd remain a use-case (all though it may be difficult to ascertain the policy file's "key" for discovered annotations - as I think the "key" is a semi-abstract value i.e. not tied to package/class name etc)... I just wonder whether we want 7.x's authorization mechanism to fallback to 6.x's - we won't get another opportunity to remove it until 8.x. Ultimately it's not my decision.

porcelli:
@manstis btw: +1 to have an unique solition; but if we collect data using annotation; @Role could be usd for such thing

porcelli:
@dgutierr i didn't understand the 2 PRs, seems that #322 has both commits

dgutierr:
@porcelli @manstis Thanks for the feedback!

It seems we have 2 main concerns:

1.- Size of the policy file.

To be honest I do not think the policy is going to be that big. May be hundreds of entries but not thousands. Anyway, the storage mechanism is encapsulated. We can always change it and split the policy into multiple files: per module, per role or whatever. 

2.- Remove old API/mechanism based on Resource/Roles

I agree on removing existing impl in favour of the new policy based mechanism. However, I'd do it incrementally, trying not to break existing stuff. We can just deprecate the old APIs and take our time to   migrate the existing stuff (droolsjpm repos basically) to the new mechanism. I do think removing existing API is too disruptive.

3.- Build the policy file from existing @Roles annotations.

I've done a full search of the @Roles annotation usage:

* kie-drools-wb    AdministrationPerspective @Roles({ "admin" })
* kie-wb               AdministrationPerspective @Roles({ "admin" })
* guvnor               M2RepoPerspective @Roles({ "admin" })
* drools-wb          M2RepoPerspective @Roles({ "admin" })
* drools-wb          AdministrationPerspective @Roles({ "admin" })
* optaplanner-wb M2RepoPerspective @Roles({ "admin" })
* optaplanner-wb AdministrationPerspective @Roles({ "admin" })
* uf-ext                 GroupsManagementPerspective @Roles({UserSystemManager.ADMIN})
* uf-ext                 UsersManagementPerspective @Roles({UserSystemManager.ADMIN})

This is only a small piece of the entire security policy the workbench defines. Right now, the there are several places where roles are assigned to resources:

* @Roles annotation: only a few perspectives are tagged this way.
* workbench-policy.properties: to map perspective menu items and a few other UI resources to roles.
* Git repo: roles assigned to Repositories and Projects

The question is: what is going to happen when we install version 7 on existing customers? 

We need to support existing security configurations whilst the new security mechanism takes over. One feasible solution could be to implement a custom WorkbenchPolicyMigrationTool that collects all the information available from all the 3 places above and populates the security policy with the proper permissions. For instance, 

* if a @Roles({"r1"}) annotation is found in a perspective "p1" then a new entry "role.r1.permission.perspective.view.p1.true" will be added to the policy. 
* If a Repository repo1 with roles "admin, analyst" is stored in git then two entries "role.admin.permission.read.repository.repo1=true" and "role.analyst.permission.read.repository.repo1=true" will be added as well. 
* For entries in the workbench-policy.properties file we just need to map (using an static mapping) the roles/features to perspective permissions.

This migration tool will be executed at startup time. Initially the security policy will be empty, and it will be populated by the migration tool only the first time the workbench is started. Once the policy is created it will take over and the only way to assign permission to resources it will be through the security policy UI editor.

There are considerable changes the workbench requires in order to support the new mechanism:

* Remove the add roles commends from the CLI tool.
* Remove the KieWorkbenchACL related classes and usage.
* Migration tool

The above is only a feasible solution. We still need to figure out whether it is doable or what are the alternatives. We can not do everything at the same time, so that's why I think it is best to go incrementally.

Having said that, I believe a mechanism for scanning the @Roles annotation and generate a policy file does not worth the effort. Moreover, if we are thinking of removing old API (including the @Roles annotation) what is the point of having such mechanism?

porcelli:
@dgutierr we're doing a major release, so I wouln't limit ourselves to do the `right thing` due potential compat issues; of course we'll need provide something for existing users to migrate to 7, but it should not limit us any potential breaks.

And related to that, I think you can execute the changes incremental as you prefer, but in order to the `rule` that we agreed for a while (just bring to master stable features), I'd keep this security model in a feature branch until we can completely replace existing solution. And there are some important changes to do, as today Resorce is just a marker interface and the real actions area around RuntimeResource, so I'd just consider this new securuty model once all the existing structure is changed.

dgutierr:
Added important changes according the feedback received:

* Legacy API has been completely removed from Uberfire codebase. Right now the authz control mechanism only relies on permissions. 

This is a big impact which breaks existing APIs and it will cause downstream projects to not compile. That means several PRs on those projects (kie-*) are required before merging this PR.

The new implementation provides an extensible fluent API to check for authorization rights either on resources or permissions.

https://github.com/dgutierr/uberfire/blob/UF-280/uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/authz/WorkbenchControllerHelper.java
https://github.com/dgutierr/uberfire/blob/UF-280/uberfire-workbench/uberfire-workbench-client/src/test/java/org/uberfire/client/authz/WorkbenchControllerTest.java

The above pattern can be extended to other resources like, repositories, projects, etc... with the aim of creating an API to let the developers deal with authz management.

* Voting strategy support

The VotingStrategy concept is still valid in the new API implementation. Now, it is always possible (and optional) to specify the voting strategy on every request passed through  the permission manager.  

https://github.com/dgutierr/uberfire/blob/UF-280/uberfire-security/uberfire-security-api/src/test/java/org/uberfire/security/impl/authz/AuthorizationManagerTest.java#L266


* Authorization policy file split. Added the changes to be able to split the policy into small modules.

https://github.com/dgutierr/uberfire/blob/UF-280/uberfire-backend/uberfire-backend-server/src/test/java/org/uberfire/backend/server/authz/AuthzPolicyStorageTest.java
https://github.com/dgutierr/uberfire/tree/UF-280/uberfire-backend/uberfire-backend-server/src/test/resources/WEB-INF/classes

dgutierr:
@psiroky Any idea why does jenkins complain?

psiroky:
Not sure if its related, but the PR has merge conflicts.

manstis:
@dgutierr I'll look more on Monday (when at a computer vs phone).. I've spoken to @porcelli, a long time ago, about logical to physical Place identifier mapping and wonder if your "home perspective" changes could be made more generic to support logical to physical mapping for any Place and any role (although I accept it may well be out of scope for these *ACL* changes).

uberfire-api/src/main/java/org/uberfire/security/Resource.java
```diff
@@ -16,9 +16,40 @@
...
+     *
+     * <p>The dependency list is used for instance to determine if a user can access a given resource. Should
+     * the access to all its dependencies is denied, it is denied for this instance as well.</p>
+     *
+     * @return A list of resources or null if this resource has no dependencies.
+     *
+     * @see AccessController
+     */
+    default List<Resource> getDependencies() {
+        return null;
```
porcelli: -1 for use of null

uberfire-api/src/main/java/org/uberfire/security/Resource.java
```diff
@@ -16,9 +16,40 @@
...
+     */
+    default String getIdentifier() {
+        return null;
+    }
+
+    /**
+     * Get the resource type classifier
+     */
+    default ResourceType getResourceType() {
+        return null;
```
porcelli: -1 for use of null

uberfire-api/src/main/java/org/uberfire/security/ResourceAction.java
```diff
@@ -0,0 +1,38 @@
...
+ * <p>This interface is intended to be extended by the different {@link Resource} types. It is up to every
+ * resource type implementation to define the list of available actions.</p>
+ */
+public interface ResourceAction {
+
+    /**
+     * The view action is common to all resource types.
+     * <p>Basically, it refers to the ability to access (view, read, ...) a resource</p>
+     */
+    ResourceAction VIEW = () -> "view";
```
porcelli: I'd use `Read` as CRUD

uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/authz/DefaultSecurityPolicyStorage.java
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    @Inject
+    public DefaultSecurityPolicyStorage(PermissionManager permissionManager) {
+        this.permissionManager = permissionManager;
+    }
+
+    @Override
+    public AuthorizationPolicy loadPolicy() {
+        URL fileURL = Thread.currentThread().getContextClassLoader().getResource("security-policy.properties");
```
porcelli: why not use VFS impl?

uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/authz/DefaultSecurityPolicyStorage.java
```diff
@@ -0,0 +1,231 @@
...
+
+            logger.info("Security policy loaded: " + fileURL.getPath());
+            return policy;
+        } else {
+            logger.info("Security policy not detected.");
+            return null;
+        }
+    }
+
+    public AuthorizationPolicy loadPolicy(File policyDir) {
```
porcelli: why not use VFS impl?

uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/authz/DefaultSecurityPolicyStorage.java
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    public boolean isPolicyFile(String fileName) {
+        return fileName != null &&
+                (fileName.equals("security-policy.properties") ||
+                        fileName.startsWith("security-module-"));
+    }
+
+    public void loadPolicyFile(AuthorizationPolicyBuilder builder, File file) {
+        try (InputStreamReader reader = new InputStreamReader(new FileInputStream(file))) {
```
porcelli: why not use VFS impl?

uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/security/FileSystemResourceAdaptor.java
```diff
@@ -48,12 +46,7 @@ public String getSignatureId() {
...
     @Override
-    public Collection<String> getGroups() {
-        return Collections.emptyList();
-    }
-
-    @Override
-    public Collection<String> getTraits() {
-        return Collections.emptyList();
+    public ResourceType getResourceType() {
+        return () -> "filesystem";
```
porcelli: why not create a default filesystem ResourceType?

uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/security/RoleLoader.java
```diff
@@ -0,0 +1,109 @@
...
+                "security-policy.properties",
+                "security-management.properties",
+                "WEB-INF/classes/security-policy.properties",
+                "WEB-INF/classes/security-management.properties"};
+
+        for (String path : paths) {
+            URL pathURL = Thread.currentThread().getContextClassLoader().getResource(path);
+            if (pathURL != null) {
+                File webInf = new File(pathURL.getPath()).getParentFile().getParentFile();
+                return new File(webInf, "web.xml");
```
porcelli: VFS?

uberfire-api/src/main/java/org/uberfire/security/Resource.java
dgutierr: Initially, default methods were added to not break api compatibility which is no longer a req. So  I'm ok with removing the default classifier and let them as pure abstract methods.

uberfire-api/src/main/java/org/uberfire/security/ResourceAction.java
dgutierr: access, view, read, ... whatever. I'll change it to READ.

uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/authz/DefaultSecurityPolicyStorage.java
dgutierr: Exactly! That's the aim. Although we also need a mechanism for deploying an initial security policy. I mean, any UF app needs a way to put the default policy file as part of the webapp artifacts so that this policy is deployed the first time the app is started. Once deployed and stored into VFS, the only way to modify the policy will be through the UI or modifying the VFS content directly.

In kie-wb, this default policy file will replace the existing workbench-policy.properties, i.e: what are the features the roles are granted/denied by default.

uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/security/FileSystemResourceAdaptor.java
dgutierr: +1

uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/security/RoleLoader.java
dgutierr: Well, roles come from web.xml descriptor. The less redundancy we have the better. So honestly, I do not really see the need for creating another registry for roles under VFS or wherever.

uberfire-api/src/main/java/org/uberfire/security/Resource.java
porcelli: so change the algorithms behind it and try use Optional, as null isn't a good return value in java 8 land

uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/authz/DefaultSecurityPolicyStorage.java
porcelli: VFS has support for standard "file://" so you'd be able to use it for git OR file system.

uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/security/RoleLoader.java
porcelli: use file:// from VFS

porcelli:
One other thing that I noticed is the traditional use of setters and gettes, and as we all know this is not much immutable friendly at best;
The potential issues on parallel computations, threads and related doesn't worry me much, what worries me a bit is the update cycle of a change using setter: when I change a value using a setter is this automatically perssisted? Is this taking effect right away?
This is not necessarly an issue to fix, but a thing to keep in mind.

porcelli:
btw: if it would be expected any update in place, I'd change data structures to concurrent ones.

uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/authz/WorkbenchController.java
```diff
@@ -0,0 +1,66 @@
...
+ * {@code Button deleteButton;
+ *   WorkbenchController workbenchController;
+ *   PerspectiveActivity perspective1;
+ *
+ *   workbenchController.perspective(perspective1).delete()
+ *     .granted(() -> deleteButton.setEnabled(true))
+ *     .denied(() -> deleteButton.setEnabled(false))
+ * }</pre>
+ */
+public interface WorkbenchController {
```
porcelli: not sure if I got the purpose of this controller, can you explain?

dgutierr: It is intended for developers who want to check access to workbench resources. Imagine for instance, you want to check if an user can read a given perspective. The WorkbenchController interface provides the minimal API you need for those kind of checks.  Basically, it wraps the AuthorizationManager adding a fluent check API on top.

It can also be taken as an example of how to build a custom controller, like for instance: 

https://github.com/dgutierr/uberfire-tutorial/blob/UF-280/uftasks/uftasks-showcase/uftasks-webapp/src/main/java/org/uberfire/shared/authz/UFTasksController.java

In kie-wb I think we will have a KIEAccessController which will be the main interface developers will use for checking access to KIE related resources (org units, repos, projects, ...)

dgutierr:
@porcelli Last commit contains all the changes you recommended.

Except the changes related to ""file://". I'm not sure if I get what you mean. Can you elaborate please?

dgutierr:
VFS API changes suggested done.

Security annotations "@PermissionCheck" & "@ResourceCheck" added.

Those annotations has some restrictions: 
* Only work on client side 
* Only on CDI managed beans
* Only can be added to methods that return "void" 
* Only work if the method is called from an external component.
* For inner component calls the solution is to use a delegate or a "self" instance.

Example usage:

https://github.com/dgutierr/uberfire-tutorial/blob/UF-280/uftasks/uftasks-showcase/uftasks-webapp/src/main/java/org/uberfire/client/screens/ProjectsPresenter.java#L84


@porcelli  Can you review please?

psiroky:
jenkins retest this

dgutierr:
I finished the permission editor UI implementation. This implies several changes:

1.- The user & group management perspectives has been melted into a single one called Security.
2.- Ability to list & edit the roles defined at web.xml.
3.- Ability to set the home perspective per role & group.
4.- Ability to edit the permissions for any role or group.
5.- All the security configuration is stored into a new internal git repo called "security".

csadilek:
I've been following along and this is starting to look really good! However, before this can be merged, we need:

- UI updates and polishing based on UXD feedback
- @porcelli to take another look and "plus one" especially in regards to the planned security changes for cloud (i.e. is this generic enough it can be merged for now and refined later?)
- @ederign to review as well as this is a lot of new code for UF :)
- Commits to be squashed into a single one

Some comments:

- RoleEditorView.ui.xml introduced in this change set is using UiBinder while the rest of the new templates are Errai UI (nice job there). Is this to keep roles/groups consistent as they have originally been implemented with UiBinder? I would definitely prefer not to introduce more UiBinder code.

- The ```data-field``` attribute is optional and ```id=``` can be used instead which makes collaboration with UXD easier (the templates can be consumed as is as long as elements have ids). The only reason to use ```data-field``` is for components that get rendered multiple times on the same page where the ids would otherwise not be unique.

dgutierr:
Thanks @csadilek for your feedback.

* UXD feedback => Indeed, I'm waiting for Joe's feedback.

* RoleEditorView.ui.xml => yes, it is.

* I'll do a 2nd review of the templates to ensure I'm using the data-field correctly.

dgutierr:
@psiroky do you know why the build is failing?

psiroky:
@dgutierr it's all there in the build log. This is just excerpt from the error message list:
```
[ERROR] Errors in 'jar:file:/home/jenkins/workspace/uberfire-pullrequests/uberfire-security/uberfire-security-client/target/uberfire-security-client-0.9.0-SNAPSHOT-sources.jar!/org/uberfire/security/processors/ResourceCheckProcessor.java'
11:05:46 [INFO]          [ERROR] Line 112: No source code is available for type org.jboss.errai.codegen.meta.MetaClass; did you forget to inherit a required module?
11:05:46 [INFO]          [ERROR] Line 137: No source code is available for type org.jboss.errai.codegen.builder.BlockBuilder<T>; did you forget to inherit a required module?
```

csadilek:
@dgutierr I think the build failures / GWT compilation problems you're seeing for production compiles is caused by this line in all showcases: https://github.com/uberfire/uberfire/blob/master/uberfire-showcase/uberfire-webapp/pom.xml#L272

This copies the source code of the corresponding jars into the project and GWT compiles it. It's only necessary for UberFire and the workbenches as there the sources are not packaged with the main jar files (unlike Errai). So, what happens is that your rebind code (decorators, generators, etc., in uberfire-security-client that are not meant to be GWT compiled) are copied in for GWT compilation and that fails.

So, unfortunately, what you would need to do is to add a separate module uberfire-security-codegen and put your decorators/processors (and an ErraiApp.properties) there. Then add this dependency to the pom but NOT to the compileSourceArtifacts list. That should fix it. //cc @mbarkley @psiroky

dgutierr:
Thanks @csadilek !

I did exactly as you mention. Now it is working as expected.

dgutierr:
jenkins retest this

csadilek:
@dgutierr you're welcome. from my point of view let's get the UI feedback in and then merge this. Just a heads-up that there is now a conflict which must be resolved with your next commit.

uberfire-api/src/main/java/org/uberfire/security/ResourceRef.java
```diff
@@ -0,0 +1,66 @@
...
+        this.type = type;
+    }
+
+    @Override
+    public List<Resource> getDependencies() {
+        return dependencies;
+    }
+
+    public void setDependencies(List<Resource> dependencies) {
+        this.dependencies = dependencies;
```
porcelli: why not clear() and addAll()?

uberfire-api/src/main/java/org/uberfire/security/ResourceRef.java
```diff
@@ -0,0 +1,66 @@
...
+    private List<Resource> dependencies = null;
+
+    public ResourceRef(String identifier, ResourceType type) {
+        this(identifier, type, null);
+    }
+
+    public ResourceRef(String identifier, ResourceType type, List<Resource> dependencies) {
+        this.identifier = identifier;
+        this.type = type;
+        this.dependencies = dependencies;
```
porcelli: why not clear() and addAll()?

uberfire-api/src/main/java/org/uberfire/workbench/model/menu/MenuItem.java
```diff
@@ -29,16 +32,34 @@
...
+    int getOrder();
 
-    public int getOrder();
+    /**
+     * Get the list of {@link ResourceActionRef} actions this menu item is restricted to.
+     *
+     * <p>The menu item will be available provided all the given actions are authorized within the current context.</p>
+     */
+    default List<ResourceActionRef> getResourceActions() {
+        return null;
```
porcelli: empty list instead of null?

uberfire-api/src/main/java/org/uberfire/workbench/model/menu/MenuItem.java
```diff
@@ -29,16 +32,34 @@
...
+    }
+
+    /**
+     * Get the list of permission names this menu item is restricted to.
+     *
+     * <p>The menu item will be available provided all the given permissions are authorized within the current context.</p>
+     */
+    default List<String> getPermissions() {
+        return null;
```
porcelli: empty list instead of null?

uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/authz/AuthorizationPolicyVfsStorage.java
```diff
@@ -0,0 +1,173 @@
...
+    private PermissionManager permissionManager;
+    private IOService ioService;
+    private FileSystem fileSystem;
+    private Path root;
+
+    public AuthorizationPolicyVfsStorage() {
+    }
+
+    @Inject
+    public AuthorizationPolicyVfsStorage(@Named("ioStrategy") IOService ioService,
```
porcelli: ioStrategy has by default indexers.. maybe better use "configIO"

uberfire-extensions/uberfire-security/uberfire-security-management/uberfire-security-management-api/src/main/java/org/uberfire/ext/security/management/api/validation/RoleValidator.java
```diff
@@ -28,10 +29,10 @@
  * @since 0.8.0
  */
 public abstract class RoleValidator implements EntityValidator<Role> {
-    
+
     @Override
     public Set<ConstraintViolation<Role>> validate(Role entity) {
-        throw new UnsupportedOperationException("Roles are not supported.");
+        Set<ConstraintViolation<Role>> result = new HashSet<>();
+        return result;
```
porcelli: why not just return new HashSet<>();

uberfire-extensions/uberfire-security/uberfire-security-management/uberfire-security-management-client-wb/src/main/java/org/uberfire/ext/security/management/client/screens/explorer/SecurityExplorerView.ui.xml
```diff
@@ -0,0 +1,42 @@
...
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<!DOCTYPE ui:UiBinder SYSTEM "http://dl.google.com/gwt/DTD/xhtml.ent">
+<ui:UiBinder xmlns:ui="urn:ui:com.google.gwt.uibinder"
+             xmlns:b="urn:import:org.gwtbootstrap3.client.ui">
+
+  <ui:with field='i18n' type='org.uberfire.ext.security.management.client.resources.i18n.UsersManagementWorkbenchConstants'/>
```
porcelli: -1 for UiBinder

uberfire-extensions/uberfire-security/uberfire-security-management/uberfire-security-management-webapp/src/main/java/org/uberfire/ext/security/management/client/acl/GeneralTreeProvider.java
```diff
@@ -0,0 +1,135 @@
...
+import org.uberfire.security.authz.PermissionManager;
+import org.uberfire.security.client.authz.tree.LoadCallback;
+import org.uberfire.security.client.authz.tree.LoadOptions;
+import org.uberfire.security.client.authz.tree.PermissionNode;
+import org.uberfire.security.client.authz.tree.PermissionTreeProvider;
+import org.uberfire.security.client.authz.tree.impl.PermissionGroupNode;
+import org.uberfire.security.client.authz.tree.impl.PermissionLeafNode;
+
+/**
+ * This is a dummy example (only for demo purposes) of how to create custom permission tree nodes
```
porcelli: why not move this to test ?

uberfire-extensions/uberfire-security/uberfire-security-management/uberfire-security-management-webapp/src/main/java/org/uberfire/ext/security/management/client/acl/UIAssetsTreeProvider.java
```diff
@@ -0,0 +1,101 @@
...
+    public PermissionNode buildRootNode() {
+        if (isActive()) {
+            PermissionGroupNode rootNode = new PermissionGroupNode(this);
+            rootNode.setPositionInTree(rootNodePosition);
+            rootNode.setNodeName(_getRooNodeName());
+            return rootNode;
+        }
+        else {
+            // Hide the node in the tree
+            return null;
```
porcelli: null :(

uberfire-extensions/uberfire-security/uberfire-security-management/uberfire-widgets-security-management/src/main/java/org/uberfire/ext/security/management/client/widgets/management/editor/acl/ACLEditor.java
```diff
@@ -0,0 +1,113 @@
...
+import org.uberfire.ext.security.management.client.widgets.management.editor.acl.node.PermissionWidgetFactory;
+import org.uberfire.security.authz.Permission;
+import org.uberfire.security.authz.PermissionCollection;
+import org.uberfire.security.client.authz.tree.PermissionNode;
+import org.uberfire.security.client.authz.tree.PermissionTree;
+import org.uberfire.security.client.authz.tree.PermissionTreeFactory;
+import org.uberfire.security.impl.authz.DefaultPermissionCollection;
+
+@Dependent
+public class ACLEditor implements IsWidget {
```
porcelli: Widget? ErraiUI allows us to avoid such thing... isn't it?

uberfire-extensions/uberfire-security/uberfire-security-management/uberfire-widgets-security-management/src/main/java/org/uberfire/ext/security/management/client/widgets/management/editor/acl/ACLSettings.java
```diff
@@ -0,0 +1,179 @@
...
+import org.uberfire.client.mvp.PerspectiveActivity;
+import org.uberfire.client.mvp.UberView;
+import org.uberfire.ext.security.management.client.widgets.management.events.HomePerspectiveChangedEvent;
+import org.uberfire.ext.security.management.client.widgets.management.events.PriorityChangedEvent;
+import org.uberfire.ext.widgets.common.client.dropdown.PerspectiveDropDown;
+import org.uberfire.security.authz.AuthorizationPolicy;
+import org.uberfire.security.authz.PermissionManager;
+
+@Dependent
+public class ACLSettings implements IsWidget {
```
porcelli: IsWidget seems not needed by ErraiUI

uberfire-extensions/uberfire-security/uberfire-security-management/uberfire-widgets-security-management/src/main/java/org/uberfire/ext/security/management/client/widgets/management/editor/acl/node/BasePermissionNodeEditor.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    @Override
+    public void setTreeLevel(int treeLevel) {
+        this.treeLevel = treeLevel;
+    }
+
+    @Override
+    public int getNodePanelWidth() {
+        return width - leftMargin - (treeLevel*padding);
```
porcelli: CSS?

uberfire-extensions/uberfire-security/uberfire-security-management/uberfire-widgets-security-management/src/main/java/org/uberfire/ext/security/management/client/widgets/management/editor/acl/node/LeafPermissionNodeEditorView.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    @Override
+    public void setNodeName(String name) {
+        nodeName.setText(name);
+    }
+
+    @Override
+    public void setNodePanelWidth(int width) {
+        nodeNamePanel.getStyle().setProperty("width", width + "px");
```
porcelli: cs

uberfire-extensions/uberfire-security/uberfire-security-management/uberfire-widgets-security-management/src/main/java/org/uberfire/ext/security/management/client/widgets/management/editor/acl/ACLSettings.java
mbarkley: To clarify, if you are using this type as `@DataField` somewhere, then it can either be an `IsWidget` or an [IsElement](https://github.com/errai/errai/blob/master/errai-common/src/main/java/org/jboss/errai/common/client/api/IsElement.java).

If you make this type `@Templated` then it need not be either.

uberfire-extensions/uberfire-security/uberfire-security-management/uberfire-widgets-security-management/src/main/java/org/uberfire/ext/security/management/client/widgets/management/editor/group/GroupEditorView.ui.xml
```diff
@@ -16,34 +16,58 @@
...
              xmlns:b="urn:import:org.gwtbootstrap3.client.ui">
 
   <ui:with field="i18n" type="org.uberfire.ext.security.management.client.resources.i18n.UsersManagementWidgetsConstants"/>
 
   <ui:style>
     .groupTitle {
       float: left;
-      margin-left: 10px;
-      margin-top: 20px;
+      margin-top: 0;
```
porcelli: -1 for UiBinder

uberfire-extensions/uberfire-security/uberfire-security-management/uberfire-widgets-security-management/src/main/java/org/uberfire/ext/security/management/client/widgets/management/editor/group/workflow/GroupEditorWorkflow.java
```diff
@@ -0,0 +1,351 @@
...
+
+import static org.uberfire.workbench.events.NotificationEvent.NotificationType.*;
+
+/**
+ * <p>Main entry point for viewing a group instance.</p>
+ * 
+ * @since 0.8.0
+ */
+@Dependent
+public class GroupEditorWorkflow implements IsWidget {
```
porcelli: Too much logic for a View; It should have a presenter

uberfire-extensions/uberfire-security/uberfire-security-management/uberfire-widgets-security-management/src/main/java/org/uberfire/ext/security/management/client/widgets/management/editor/role/RoleEditorView.ui.xml
```diff
@@ -0,0 +1,69 @@
...
+  ~  
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<!DOCTYPE ui:UiBinder SYSTEM "http://dl.google.com/gwt/DTD/xhtml.ent">
+<ui:UiBinder xmlns:ui="urn:ui:com.google.gwt.uibinder"
```
porcelli: -1 uibinder

uberfire-extensions/uberfire-security/uberfire-security-management/uberfire-widgets-security-management/src/main/java/org/uberfire/ext/security/management/client/widgets/management/editor/role/workflow/BaseRoleEditorWorkflow.java
```diff
@@ -0,0 +1,270 @@
...
+
+import static org.uberfire.workbench.events.NotificationEvent.NotificationType.*;
+
+/**
+ * <p>The workflow for editing a role.</p>
+ * <p>It links the editor & sub-editors components with the editor driver and the remote user services.</p>
+ * 
+ * @since 0.9.0
+ */
+public abstract class BaseRoleEditorWorkflow implements IsWidget {
```
porcelli: lots of logic, too much dependencies

uberfire-extensions/uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/dropdown/LiveSearchDropDown.java
```diff
@@ -0,0 +1,214 @@
...
+
+    View view;
+    int maxItems = 10;
+    LiveSearchService searchService = null;
+    boolean searchEnabled = true;
+    boolean searchCacheEnabled = true;
+    Map<String,List<String>> searchCache = new HashMap<>();
+    String selectedItem = null;
+    String lastSearch = null;
+    String searchHint = "Search items ...";
```
porcelli: i18n?

uberfire-extensions/uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/dropdown/LiveSearchDropDownView.java
```diff
@@ -0,0 +1,196 @@
...
+    LiveSearchDropDown presenter;
+
+    @Override
+    public void init(LiveSearchDropDown presenter) {
+        this.presenter = presenter;
+    }
+
+    @Override
+    public void setMaxHeight(int maxHeight) {
+        dropDownMenu.getStyle().setProperty("max-height", maxHeight + "px");
```
porcelli: CSS?

uberfire-extensions/uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/dropdown/PerspectiveDropDown.java
```diff
@@ -0,0 +1,154 @@
...
+    @Override
+    public Widget asWidget() {
+        return liveSearchDropDown.asWidget();
+    }
+
+    LiveSearchService searchService = (pattern, maxResults, callback) -> {
+
+        List<String> result = new ArrayList<>();
+        for (SyncBeanDef<PerspectiveActivity> beanDef : iocManager.lookupBeans(PerspectiveActivity.class)) {
+            PerspectiveActivity p = beanDef.getInstance();
```
porcelli: better use UF API, this is going to be bad in the future

uberfire-api/src/main/java/org/uberfire/security/ResourceRef.java
dgutierr: fixed

uberfire-api/src/main/java/org/uberfire/security/ResourceRef.java
dgutierr: fixed

uberfire-api/src/main/java/org/uberfire/workbench/model/menu/MenuItem.java
dgutierr: fixed

uberfire-api/src/main/java/org/uberfire/workbench/model/menu/MenuItem.java
dgutierr: fixed

uberfire-extensions/uberfire-security/uberfire-security-management/uberfire-security-management-api/src/main/java/org/uberfire/ext/security/management/api/validation/RoleValidator.java
dgutierr: fixed

uberfire-extensions/uberfire-security/uberfire-security-management/uberfire-security-management-client-wb/src/main/java/org/uberfire/ext/security/management/client/screens/explorer/SecurityExplorerView.ui.xml
dgutierr: Well, I agree every new component we do should be Errai UI. This is a special case though since it is a refactoring of an existing functionality and I didn't want to spend extra time moving to Errai UI. So I'd leave this as is in order to keep consistency with existing implementation.

uberfire-extensions/uberfire-security/uberfire-security-management/uberfire-security-management-webapp/src/main/java/org/uberfire/ext/security/management/client/acl/GeneralTreeProvider.java
dgutierr: Well, this is part of the showcase, but is disabled by default. So it is only intended for development/demo purposes as a way of showing how to extend/customize the permission tree. I'd just leave it as that.

uberfire-extensions/uberfire-security/uberfire-security-management/uberfire-security-management-webapp/src/main/java/org/uberfire/ext/security/management/client/acl/UIAssetsTreeProvider.java
dgutierr: fixed. PermissionTreeProvider#isActive is the default way to know if a provider is ready to supply a root node.

uberfire-extensions/uberfire-security/uberfire-security-management/uberfire-widgets-security-management/src/main/java/org/uberfire/ext/security/management/client/widgets/management/editor/acl/node/BasePermissionNodeEditor.java
dgutierr: Those are dynamic calculated properties among all the nodes within the tree. So CSS is not an option here I think.

uberfire-extensions/uberfire-security/uberfire-security-management/uberfire-widgets-security-management/src/main/java/org/uberfire/ext/security/management/client/widgets/management/editor/acl/node/LeafPermissionNodeEditorView.java
dgutierr: it is related with the above. See LeafPermissionNodeEditor:85

uberfire-extensions/uberfire-security/uberfire-security-management/uberfire-widgets-security-management/src/main/java/org/uberfire/ext/security/management/client/widgets/management/editor/group/GroupEditorView.ui.xml
dgutierr: Same reason given above (see SecurityExplorerView.ui.xml)

uberfire-extensions/uberfire-security/uberfire-security-management/uberfire-widgets-security-management/src/main/java/org/uberfire/ext/security/management/client/widgets/management/editor/group/workflow/GroupEditorWorkflow.java
dgutierr: Same reason given above (see SecurityExplorerView.ui.xml).

These classes are based on existing user&group managament implementation. I would't spend too much time now. We can discuss later if a refactor to Errau UI is required.

uberfire-extensions/uberfire-security/uberfire-security-management/uberfire-widgets-security-management/src/main/java/org/uberfire/ext/security/management/client/widgets/management/editor/role/RoleEditorView.ui.xml
dgutierr: Same reason given above.

uberfire-extensions/uberfire-security/uberfire-security-management/uberfire-widgets-security-management/src/main/java/org/uberfire/ext/security/management/client/widgets/management/editor/role/workflow/BaseRoleEditorWorkflow.java
dgutierr: Same reasons given above.

uberfire-extensions/uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/dropdown/LiveSearchDropDown.java
dgutierr: fixed

uberfire-extensions/uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/dropdown/LiveSearchDropDownView.java
dgutierr: Not sure as the those properties can be changed by clients.

dgutierr:
* Added most of the changes suggested by @porcelli (not all, see comments on each file).
* Rebased master

@ederign Ready for a review?

ederign:
Hi David, great work.

ActivityBeansInfo should be used in PerspectiveDropDown in order to get perspectives ids;
You also should remove the manual lookup from  perspective and create a util class for this, like ActivityBeans Info. (Please point Workbench.java and the showcase entry point lookup for this class).
Maybe this class could also  return a perspective given an id.

Errai UI is a requirement for new widgets. You must use it for new things and maybe refactoring existing widgets in order to use it.

Did UX team already approved the new widgets?

My only concern with your solution is the overlap of some areas with keycloak. Probably in near future we will have to remove part of your solution in order to integrate with keycloak.

And thanks again for the docs!

+1 after the refactorings and uxd approval.

dgutierr:
Applied the latest fixes & recommendations made by the UXD team. 

A demo preview => https://youtu.be/4y9s89JoxXI

Waiting for UXD verification before merging.

lclay2:
+1

PR #323, opened 2016-03-15
UF-287: Added getOrFail method to encapsulate error logic in ConfigProperties class.

romartin:


uberfire-commons/src/main/java/org/uberfire/commons/config/ConfigProperties.java
```diff
@@ -82,6 +82,22 @@ public ConfigProperty get( String name, String defaultValue ) {
...
     /**
+     * Returns the ConfigProperty instance corresponding to the configured value of the given property, or the default
+     * if no configured value exists.
+     * If the property does not exist or its value is empty, it throws an unchecked runtime exception.
+     *
+     * @param name The property name. Must not be null.
+     * @param defaultValue The value to use if no configured value exists. May be null.
+     * @return The config property instance.
+     */
+    public ConfigProperty getOrFail( String name, String defaultValue ) {
```
csadilek: @romartin Why provide a default value, if you're gonna fail anyway if the property is not set?

csadilek:
API issue above, plus this needs a unit test.

uberfire-commons/src/main/java/org/uberfire/commons/config/ConfigProperties.java
ederign: it's just a POV, but I think API's that fail like that is unpolite. Why not return null and let the client decides what did with that?

uberfire-commons/src/main/java/org/uberfire/commons/config/ConfigProperties.java
```diff
@@ -92,7 +108,7 @@ public ConfigProperty get( String name, String defaultValue ) {
      *            a line of text to print before listing the configuration values
      */
     public String getConfigurationSummary( String heading ) {
-        final String newLine = System.getProperty( "line.separator" );
+        final String newLine = System.getProperty("line.separator");
```
ederign: how about if there is on line separator property? is that default in all containers?

uberfire-commons/src/main/java/org/uberfire/commons/config/ConfigProperties.java
```diff
@@ -165,4 +181,14 @@ public String toString() {
         }
     }
 
+    protected static boolean isConfigPropertySet(ConfigProperties.ConfigProperty property) {
+        if (property == null) return false;
```
ederign: this is not our code style, please add {}

romartin:
@csadilek @ederign Yeah guys does not adds nothing interesting this method.. lemme just close the PR and discard it :)

PR #329, opened 2016-03-22
Upgrade jboss-ip-bom from 6.0.2.Final to 6.0.4.Final

mbiarnes:


psiroky:
+1, merged. I've also added the repo `https://repository.jboss.org/nexus/content/groups/public/` into top-level POM so that we don't have to wait for the new release to get synced into Maven Central (this is the reason why the build failed).

PR #330, opened 2016-03-23
JBPM-5014 - OnResize needs to be called after animation finishes to a…

cristianonicolai:
…llow correct width and height


manstis:
Shouldn't the test verify onResize() *is* called - wasn't that the whole purpose of the change?

cristianonicolai:
so the answer is yes and no :) The change makes sure that onResize() is only called after the animation is finished. I created two other checks that ensure onResize() is called when onComplete() callback is executed. Had to do a bit of refactor to allow the internal inner classes to be available for testing.

manstis:
Thanks @cristianonicolai +1 to merge.

csadilek:
0.8.x: 313653fe407d014d5e02be1a99f6fb7695ca9d7d

PR #343, opened 2016-04-08
BZ1323572: AsyncWatchService dies when a RuntimeException is thrown f…

tkobayas:
…rom IOWatchServiceExecutorImpl.execute()

Confirmed that it fixes the issue in BRMS 6.2.2.

### DIFF
--- a/uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/io/watch/AbstractIOWatchService.java
+++ b/uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/io/watch/AbstractIOWatchService.java
@@ -157,7 +157,11 @@ public abstract class AbstractIOWatchService implements IOWatchService,
                         break;
                     }
 
-                    wsExecutor.execute( wk, AbstractIOWatchService.this );
+                    try {
+                        wsExecutor.execute( wk, AbstractIOWatchService.this );
+                    } catch ( final Exception ex ) {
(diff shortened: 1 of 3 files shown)


manstis:
@tkobayas I see the fix is pretty straight forward; but could you please write a Unit Test too (ideally including execution of a ```WatchService``` from ```AbstractIOWatchService.addWatchService(...)``` but since this might either require extensive mocking or even a code restructure I guess simply showing Exceptions in ```AsyncWatchService.execute(...)``` are handled would be fine).

BTW, just an opinion, but would the ```try {...} catch(..)``` block be better positioned around https://github.com/tkobayas/uberfire/blob/bz1323572/uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/io/watch/AbstractIOWatchService.java#L160 IMO this makes the code more readable. WDYT?

tkobayas:
Hi Mike, thank you for the comment.

I agree that try-catch in AbstractIOWatchService.java is better. Also I added a unit test. In order to avoid modifying the product codes, I used mock, system property (to disable autostart) and reflection (to get an asyncWatchService) in the unit test. Does it look okay?

manstis:
@tkobayas Looks fine; thank-you.. just a couple of comments:-

1) Please add the license header to the test class
2) Could you please squash the commits into a single commit.

tkobayas:
Sure, done!

manstis:
@tkobayas Thanks.

+1 to merge.

ederign:
@tkobayas thanks!

PR #344, opened 2016-04-08
[UF-299] disable git/ssh daemons by default when running tests

psiroky:
* disabling the git/ssh daemons by default makes the tests more stable.
   Maven build is usually executed in parallel (e.g. -TC), which means
   tests in multiple modules are executed in parallel (in different JVM of course).
   That sometimes leads into random test failures caused by failure
   to bind to the default git/ssh ports.

Tested with both uberfire and uberfire-extensions repos (and with kie-wb running to make sure the ports are already bound).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -259,6 +259,11 @@
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-surefire-plugin</artifactId>
           <configuration>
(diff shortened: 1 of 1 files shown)


ederign:
+1

porcelli:
+1

ederign:
@psiroky thanks!

PR #345, opened 2016-04-08
[UF-299] disable git/ssh daemons by default when running tests

psiroky:
* disabling the git/ssh daemons by default makes the tests more stable.
   Maven build is usually executed in parallel (e.g. -TC), which means
   tests in multiple modules are executed in parallel (in different JVM of course).
   That sometimes leads into random test failures caused by failure
   to bind to the default git/ssh ports.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -289,6 +289,11 @@
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-surefire-plugin</artifactId>
           <configuration>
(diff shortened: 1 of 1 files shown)


psiroky:
I know 0.8.x is frozen now, but this is only test/build fix. @etirelli I believe it can be pushed, can you confirm?

etirelli:
Yes, if it only fixes tests, then it is fine to merge.

PR #347, opened 2016-04-11
Zanata-translation-changes 2016.04.11

mbiarnes:


manstis:
@mbiarnes Hi.. just want to check you're OK with the license headers being removed (knowing the pain you've gone through to get them corrected).

ederign:
@mbiarnes could you please take a look on this?

mbiarnes:
@manstis: yes - of course - if this is not needed in this files we should get rid of them
+1 to remove useless license headers

manstis:
@mbiarnes hehehehe.. personally I don't mind whether they're there or not; but you had lots of problems correcting *some* license headers; so assume *you* wanted the licenses to be present. This PR deletes some headers so I was asking if you're OK with that..., since you've +1'ed then.. +1 from me too.

PR #349, opened 2016-04-13
fix some exceptions and exception messages (mostly adding messages)

zkrejcov:
See also:
https://github.com/droolsjbpm/drools-wb/pull/162
https://github.com/droolsjbpm/kie-uberfire-extensions/pull/26

### DIFF
--- a/uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/security/JAASAuthenticationService.java
+++ b/uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/security/JAASAuthenticationService.java
@@ -77,7 +77,7 @@ public class JAASAuthenticationService extends GroupAdapterAuthorizationSource i
 
             return user;
         } catch ( final LoginException ex ) {
-            throw new FailedAuthenticationException();
+            throw new FailedAuthenticationException("Failed to authenticate user " + username, ex);
         }
     }
 
--- a/uberfire-nio2-backport/uberfire-nio2-api/src/main/java/org/uberfire/java/nio/file/Files.java
+++ b/uberfire-nio2-backport/uberfire-nio2-api/src/main/java/org/uberfire/java/nio/file/Files.java
@@ -1091,13 +1091,13 @@ public final class Files {
             }
             return result;
         } catch ( java.io.IOException ex ) {
-            throw new IOException( ex );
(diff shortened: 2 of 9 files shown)


manstis:
@porcelli Could you please review?

jhrcek:
Just a comment about these "exception message adding" commits to raise the awareness of the general problem we're sometimes facing.

Given that error handling in many backend service calls consists of only displaying the exception message in modal dialog (typical example is [here](https://github.com/droolsjbpm/jbpm-console-ng/blob/master/jbpm-console-ng-human-tasks/jbpm-console-ng-human-tasks-client/src/main/java/org/jbpm/console/ng/ht/client/editors/tasklogs/TaskLogsPresenter.java#L98) or using [DefaultErrorCallback](https://github.com/uberfire/uberfire-extensions/blob/master/uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/callbacks/DefaultErrorCallback.java) it is very important to use informative error messages to avoid situation when modal with message like "NullPointerException: null" is displayed and the exception (and stacktrace) is otherwise lost. This leads to difficulties when steps to reproduce the failure are not clear.

@porcelli @manstis Shoudln't we for example improve the DefaultErrorCallback so that the stacktrace is logged to browser javascript console or something so that we're not loosing valuable failure information? Or are there some other options  to handle this that I'm not aware of?

manstis:
@jhrcek I'm not sure what value the callstack is in JS without "source maps" enabled (which is disabled in builds other than for SuperDevMode, i.e. development). I fear a callstack would be a hard-to-read string of obfuscated method calls. @csadilek Do you have any experience of this?

zkrejcov:
@jhrcek +1 to informative exception messages, definitely. Actually to proper exception handling in general.. ;)

csadilek:
@manstis @jhrcek If I understand the question correctly, symbol obfuscation isn't an issue here as this is about server-side exceptions (and details) getting serialized to the client. However, I would decide against this as you don't want to leak that level of server-side internals to the client. IMO, it's sufficient to provide a meaningful error message to the user and, at the same time, log the details (i.e. the stack trace) on the server. //cc @porcelli @zkrejcov

manstis:
+1 to merge.

porcelli:
just a comment, most of NIO2 backport should be removed as we're already on Java 8

PR #351, opened 2016-04-14
UF-300: bug goTo(screen) doesn't work properly on JS Perspectives

ederign:
**Description**: bug goTo(screen) doesn't work properly on JS Perspectives. This can be reproduced in uberfire showcase, using the screens menu to load screens in default perspective.

**Cause**: The root cause os this issue is this:
 https://github.com/uberfire/uberfire/blob/1c14cbfd947da6f4dd8a76c062295186dee9b4b9/uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/workbench/panels/impl/AbstractSimpleWorkbenchPanelView.java#L90

All the presenters that support only one widget ( SimpleWbPresenter, ClosablePresenter,SimpleDndPresenter) have this base view. 

The issue happens when one of this panels is the root of a perspective and we do a goto(screen). Then this screen will be launched at the root panel and this view silently ignored attaching the view breaking the workbench because Uberfire cannot close everything.

The same behavior was well implemented in the StaticWorkbenchPanelView and UnanchoredStaticWorkbenchPanelview : 

https://github.com/uberfire/uberfire/blob/a0bf142be9c6cc63704e0de0e13d6ea5ec460d81/uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/workbench/panels/impl/StaticWorkbenchPanelView.java#L98

When an activity is launched in this panel, they do the proper cleanup calling place manager to close the activity.

**Solution**:

Before adding any part in this type of presenters, I removed the current part in a proper way (i.e. ClosableSimpleWorkbenchPanelPresenter) to prevent part leak. I also refactored StaticWorkbenchPanelView and UnanchoredStaticWorkbenchPanelview logic to the respective presenter and add a lot of tests.

ps. AdaptivePanels are special one, because they behave like a Simple Panel until receives other part.

### DIFF
--- a/uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/listbar/ListBarWidgetImpl.java
+++ b/uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/listbar/ListBarWidgetImpl.java
@@ -15,46 +15,17 @@
  */
 package org.uberfire.client.views.pfly.listbar;
 
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.LinkedHashSet;
-import java.util.List;
-import java.util.Map;
-import javax.annotation.PostConstruct;
-import javax.enterprise.context.Dependent;
-import javax.enterprise.inject.Instance;
-import javax.inject.Inject;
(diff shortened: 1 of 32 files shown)


csadilek:
+1

PR #359, opened 2016-05-02
upgrade to jboss-ip-bom 7.0.0.CR2

mbiarnes:
when upgrading to jboss-ip-bom 7.0.0.CR2 uberfire (master) builds successful but two tests fail:

https://gist.github.com/mbiarnes/de29c600c020b1532849910ae57f1d08
https://gist.github.com/mbiarnes/dde61c4a0119bfedf006b93bcb43b686


psiroky:
Replaced by https://github.com/uberfire/uberfire/pull/367. I removed some of the deps already coming from ip-bom.

PR #363, opened 2016-05-05
Maven parent and BOM fixes after uf+uf-ext merge

psiroky:
* uberfire-extensions-bom and uberfire-extensions-deps
   modules were removed as they are no longer needed.
   Their content was moved into uberfire-bom and
   uberfire-parent-with-dependencies


csadilek:
+1

PR #365, opened 2016-05-06
Add simple script for updating version during release

psiroky:
@mbiarnes could you please check this?

If you agree I think we should move most of the release logic into the scripts. The Jenkins should orchestrate the calling of the scripts (instead of having the logic in the job configs). The benefit of having the scripts in the git repo is that we can easily test them locally when doing changes.


mbiarnes:
+1

PR #366, opened 2016-05-06
BZ1007859 - Add 'select all' checkbox to Process instances. Allow che…

nmirasch:
…ckBox header and not to show it at the popup. Adjust columns width calculation when they have no width defined


cristianonicolai:
+1 thanks @nmirasch

PR #367, opened 2016-05-06
Upgrade ip-bom from 7.0.0.CR1 to 7.0.0.CR2 + related fixes

psiroky:
This replaces https://github.com/uberfire/uberfire/pull/359.


csadilek:
I think we should run SDM (both mvn gwt:run and from IDEA) before merging this as we've upgraded the Jetty version in 7.0.0.CR2 anticipating the next GWT release. Just to make sure this doesn't break dev move with our current GWT version. It likely won't because in the current GWT Jetty is still packaged as part of the gwt-dev.jar, but I'd like to make sure.

psiroky:
I did try `mvn clean gwt:run` inside uberfire-showcase/uberfire-webapp -- everything went well. But it would be good if someone could try from IDEA as well.

csadilek:
+1

PR #368, opened 2016-05-09
added missing transitive dependencies

mbiarnes:


uberfire-backend/uberfire-backend-cdi/pom.xml
```diff
@@ -47,6 +47,22 @@
       <artifactId>jboss-ejb-api_3.1_spec</artifactId>
       <scope>provided</scope>
     </dependency>
+
+    <!-- illegal transitive dependencies -->
```
psiroky: I think we should not add comments like this. The comment kind of suggests that the dependencies below are illegal, which is not the case -- they were added just because they were illegally brought in transitively, instead of depending on them directly (I would vote for removing the comment).

csadilek:
I am not happy with ~600 lines of more pom.xml :(. What's the advantage of this?

psiroky:
Good practice is _not_ to rely on transitive dependencies where you actually depend directly on particular jar. It may very well happen that the artifact you depend on removes one of the dependencies you rely on (this is the transitive dependency) and then build fails with no apparent reason. 

Example:
 * uberfire-api depends on slf4-api
 * uberfire-client depends on uberfire-api and also slf4-api, but in pom.xml only declares dependency on uberfire-api (and thus relies on transitive availability of slf4j-api through uberfire-api).
 * when for some reason uberfire-api remove dependency on slf4j-api (or uberfire-client removes dependency on uberfire-api), the build will fail because uberfire-client is now missing slf4j-api

Hopefully the above explain this a bit...

csadilek:
@psiroky Sure, I can see this for external dependencies but for our own esp. the uberfire modules?

psiroky:
@csadilek I never really differentiate between internal and external dependencies. I guess it is easier to fix the possible issue if you are doing the change in "internal" dependency. In any case, if you list all the deps, it is easier for me to e.g. fix things as I can immediately see (just by looking at the POM) what the module depends. But this may just not be a good enough reason for the change.

I will, however, leave the decision to you/Eder. If you decide to not list all of the deps, then we should remove the enforcer rule as well.

csadilek:
@psiroky I wouldn't do it for dependencies to our own modules (modules part of same repo) mainly because of the pom.xml increase and because I don't quite see the value. What do others think?

psiroky:
We need to figure out what to do with this one. @csadilek would you agree with adding the direct dependencies on non-UF modules? We should be able to exclude `org.uberfire` classes from the enforcer check (https://github.com/ImmobilienScout24/illegal-transitive-dependency-check):
```
<regexIgnoredClasses>
    <regexIgnoredClass>org\.uberfire\..+</regexIgnoredClass>
</regexIgnoredClasses>
```

csadilek:
@psiroky yes, I would agree with that ;)

psiroky:
@mbiarnes could you adjust the PR according to the above comment?

mbiarnes:
@psiroky 
adding
      &lt;regexIgnoredClasses&gt;
       &nbsp;&nbsp;&lt;regexIgnoredClass&gt;org\.uberfire\..+&lt;/regexIgnoredClass&gt;
      &lt;/regexIgnoredClasses&gt;
doesn't work, org.uberfire classes still show up

mbiarnes:
@psiroky @csadilek 
done

pom.xml
```diff
@@ -131,6 +130,11 @@
               <artifactId>extra-enforcer-rules</artifactId>
               <version>1.0-beta-4-jbossorg-1</version>
             </dependency>
+            <dependency>
+              <groupId>de.is24.maven.enforcer.rules</groupId>
+              <artifactId>illegal-transitive-dependency-check</artifactId>
+              <version>1.7.4</version>
+            </dependency>
```
psiroky: This should not be needed as it is already present in ip-parent: https://github.com/jboss-integration/jboss-integration-platform-bom/blob/master/pom.xml#L383. Or does the build fail without this?

pom.xml
```diff
@@ -211,6 +215,22 @@
...
+              <goals>
+                <goal>enforce</goal>
+              </goals>
+              <configuration>
+                <rules>
+                  <illegalTransitiveDependencyCheck implementation="de.is24.maven.enforcer.rules.IllegalTransitiveDependencyCheck">
+                    <regexIgnoredClasses combine.children="append">
+                      <regexIgnoredClass>org\.uberfire\..+</regexIgnoredClass>
+                    </regexIgnoredClasses>
+                    <listMissingArtifacts>true</listMissingArtifacts>
```
psiroky: I would probably remove this as the documentation contains quite a big warning: `Caution: This option is really slow!`

psiroky: On the other hand, it provides very useful info and should hopefully influence the build time only when there ar missing artifacts (in which case it is fine). @mbiarnes could you please quickly measure what is the different in build time when you enable and disable this option?

pom.xml
```diff
@@ -211,6 +210,21 @@
...
+            <execution>
+              <id>enforce-direct-dependencies</id>
+              <goals>
+                <goal>enforce</goal>
+              </goals>
+              <configuration>
+                <rules>
+                  <illegalTransitiveDependencyCheck implementation="de.is24.maven.enforcer.rules.IllegalTransitiveDependencyCheck">
+                    <regexIgnoredClasses combine.children="append">
+                      <regexIgnoredClass>org\.uberfire\..+</regexIgnoredClass>
```
psiroky: Hmm, I am not sure , but it seems this does not work as expected? No dependencies were added and the build is not failing -- afaik at least some Errai classes should be marked as "illegal". The exclude regex probably catches also the classes under test -- e.g. all the classes in this repo. I am not sure we can actually configure the plugin in a way we want.

PR #369, opened 2016-05-10
Use javax.inject.Inject instead of com.google.inject.Inject

psiroky:
* the import was wrong, all other classes are using
   also javax.inject.Inject


manstis:
+1 to merge.

PR #370, opened 2016-05-11
uberfire-parent directly inherits from ip-bom for easier dependency overrides

psiroky:
* now we can easily override the dependency version just via property
   (since uberfire-parent extends ip-bom instead of importing it in depMgmt)

 * user BOM 'uberfire-bom' can not extend the uberfire-parent
   as that would mean polluting it with hundreds of different
   versions coming from the ip-bom -- which is not desirable
   for user BOMs

 * uberfire-parent-metadata was renamed to just uberfire-parent.
   uberfire-parent-with-dependencies was removed completely
   as it is no longer required. All UF modules should (transitively)
   use the uberfire-parent (except for the user BOM(s))

@csadilek, @ederign please take a look. This is I believe very similar to what Errai does. It should greatly ease the pain connected with overriding dependencies when needed. cc @ge0ffrey


csadilek:
+1

pom.xml
```diff
@@ -636,11 +644,309 @@
...
+        <artifactId>errai-navigation</artifactId>
+        <version>${version.org.jboss.errai}</version>
+      </dependency>
+
+      <dependency>
+        <groupId>org.jboss.errai</groupId>
+        <artifactId>errai-otec</artifactId>
+        <version>${version.org.jboss.errai}</version>
+      </dependency>
+
```
csadilek: @psiroky @ederign  just saw this now. This module does no longer exist and can be removed.

PR #373, opened 2016-05-18
UF-306: AbstractWorkbenchEditorActivity: Allow Editors to provide their own LockManager handling

manstis:
See https://issues.jboss.org/browse/UF-306

```LockManager``` in ```AbstractWorkbenchEditorActivity``` is only instantiated *if* Uberfire is handling the locking; otherwise locking is left entirely to the editor itself. I also made a few changes to ```LockTarget``` (to support simplified Unit Testing) and related changes.


manstis:
@csadilek Would you be so kind as to please review this PR?

csadilek:
+1

PR #375, opened 2016-05-18
UF-308: UpdatedLockStatusEvent: Include Path parameter.

manstis:
See https://issues.jboss.org/browse/UF-308

This PR simply adds the ```Path``` associated with the ```UpdatedLockStatusEvent``` so I can disambiguate events relating to different locks.


csadilek:
+1

PR #376, opened 2016-05-19
add test for ValidationUtils

zkrejcov:
(part X of moving-tests-to-community series)


yurloc:
Nice! I only miss one more positive example for "hidden" files, e.g. `.gitignore`.

zkrejcov:
@yurloc afaik, those we actually do not allow and do not want to allow as valid filenames - I'm sure I could dig up some old Bugzilla for that...

yurloc:
I see. If the purpose of the class is to validate user input only and if users are not allowed to create hidden files, then it makes sense.

manstis:
+1 to merge.

Sweet test @zkrejcov :+1:

PR #386, opened 2016-06-10
RHBPMS-4014 Tasks fails on load for Chinese localization

nmirasch:


manstis:
+1 to merge.

uberfire-extensions/uberfire-widgets/uberfire-widgets-commons/src/test/java/org/uberfire/ext/widgets/common/client/common/DatePickerTest.java
```diff
@@ -56,6 +53,7 @@ public void testSetFormat() {
         String gwtDateFormat = "dd-MMM-yyyy";
         DateTimeFormat gwtDateTimeFormat = DateTimeFormat.getFormat(gwtDateFormat);
 
+        datePicker.setLocaleName("en");
```
ederign: @nmirasch the case is sensitive? I saw your changes above changing the case of en to EN

PR #387, opened 2016-06-10
GUVNOR-2507: Guided Decision Table Editor: Add ability to open multiple decision tables in the same view

manstis:
See https://issues.jboss.org/browse/GUVNOR-2507

This PR adds a ```UberView<P>``` equivalent for use with ```Element``` based Views and allows me to override a couple of bits in ```LockManagerImpl``` in a subclass I have in ```drools-wb``` (where I only want to fire the ```UpdatedLockStatusEvent``` if the lock has changed for the visible/active document - tied to opening multiple decision tables in the same editor).

### DIFF
--- a/uberfire-client-api/src/main/java/org/uberfire/client/mvp/UberElement.java
+++ b/uberfire-client-api/src/main/java/org/uberfire/client/mvp/UberElement.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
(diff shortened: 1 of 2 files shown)


manstis:
Part of an ensemble:-

https://github.com/uberfire/uberfire/pull/387
https://github.com/droolsjbpm/kie-wb-common/pull/382
https://github.com/droolsjbpm/drools-wb/pull/194

csadilek:
+1

PR #388, opened 2016-06-10
RHBPMS-4014 Tasks fails on load for Chinese localization

nmirasch:
@manstis @ederign  can you take a look? I have changed the code to avoid to use  the framework default language String. Only sets the datePicker to DatePickerLanguage.EN when the locale is "default" or it is not defined.


manstis:
@nmirasch Looks OK to me... @ederign OK with you now?

ederign:
@nmirasch thanks!

PR #389, opened 2016-06-13
Upgrade wildfly-dist to 10.0.0.Final.

psiroky:
* refactored the wildfly version property
   into uf-parent for better reusability

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -62,7 +62,9 @@
     <version.org.jboss.keycloak>1.9.0.Final</version.org.jboss.keycloak>
     <version.org.patternfly>2.1.0</version.org.patternfly>
     <version.org.picketlink>2.6.0.Final</version.org.picketlink>
-    <version.org.wildfly>8.2.0.Final</version.org.wildfly>
+    <!-- WildFly version used together with the GWT's Super Dev Mode -->
+    <version.org.wildfly.gwt.sdm>10.0.0.Final</version.org.wildfly.gwt.sdm>
+    <version.org.wildfly.core>2.1.0.Final</version.org.wildfly.core>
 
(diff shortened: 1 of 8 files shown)


csadilek:
Needs https://github.com/errai/errai/pull/176 to be merged first. The test failure is also related, unfortunately. I can look into that in a bit. //cc @mbarkley @psiroky

mbarkley:
@psiroky @csadilek 

It looks like the problem is caused by a bug in the wildfly-domain-management artifact for the new version (2.1.0.Final). You can see the problematic code [here](https://github.com/wildfly/wildfly-core/blob/2.1.0.Final/domain-management/src/main/java/org/jboss/as/domain/management/security/PropertiesFileLoader.java#L230-L238).

The failing test tries to delete a user. But when this persistProperties method is called, it still writes the line with the deleted user to the properties file in the else block. It looks like this problem exists in all the newer versions (2.2.0.CR2 and 3.0.0.Alpha1) as well.

I can send them a PR to fix this later today, but how will we proceed in the interim?

csadilek:
@psiroky we have a fix for this on my fork (amended your commit): https://github.com/csadilek/uberfire/commit/d5614f784f925faabad57286dd04cd81af34e894

Can you replace your commit with this one and update this PR? This is a temporary solution until the WF bug is fixed, but it's not worth waiting for that at this point.

psiroky:
@csadilek, @manstis  thanks guys, I will update the PR shortly.

manstis:
@psiroky I take no credit.. thank @mbarkley :)

psiroky:
@manstis right, sorry @mbarkley :) the thanks should really go to you

csadilek:
+1

PR #395, opened 2016-06-15
Moving IOCUtil from kie-wb-common to uberfire

paulovmr:


csadilek:
@paulovmr it might make sense to move this to Errai. Wdyt @mbarkley ? This could be similar to the instance provider but instead of injecting Instance<T> we'd inject something like a ContextualInstance<T> ....

csadilek:
@mbarkley nice, like the idea. the newInstance vs getInstance issue definitely points to a bug in the workbench. We'd have to go through the linked PR and make sure there's no application scoped bean involved and if there is we'd have to change it to dependent scope when applying the fix.

csadilek:
Linked PR: https://github.com/droolsjbpm/kie-wb-common/pull/386

paulovmr:
Hi @csadilek @mbarkley , I agree, it makes more sense moving this to Errai.

@csadilek Wdyt about us closing this PRs and opening a new ticket on JIRA to take care of this?

csadilek:
@paulovmr sounds good, unless this refactoring was needed for something right away, we can of course merge this first. If not, let's put this into Errai.

paulovmr:
@csadilek No, it's not needed right now, we can wait for this to be put into Errai. Do you mind creating the ticket? I think you could add more details since you are more familiar with Errai's particularties. :)

csadilek:
@paulovmr ok @mbarkley created a jira: https://issues.jboss.org/browse/ERRAI-961

PR #396, opened 2016-06-16
GUVNOR-2562: Guided Decision Table Editor: Clicking on GridLayer when editing cell does not update View

manstis:
See https://issues.jboss.org/browse/GUVNOR-2562


csadilek:
+1

PR #399, opened 2016-06-17
ERRAI-963: Remove errai-js.

mbarkley:
[This issue](https://issues.jboss.org/browse/ERRAI-963) describes why errai-js is being removed. It seems as though it was only ever referenced in Uberfire in dependency management.


ederign:
+1 thanks!

PR #405, opened 2016-06-21
UF-321: Error reading roles from web.xml and loading security policy files

dgutierr:
Fix + tests

@csadilek Can you verify please? Specially the usage of a ServletContextListener to capture the webapp init/detroy events.

### DIFF
--- a/uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/security/RoleLoader.java
+++ b/uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/security/RoleLoader.java
@@ -17,7 +17,6 @@
 package org.uberfire.backend.server.security;
 
 import java.io.File;
-import java.net.URL;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
@@ -32,6 +31,8 @@ import org.jdom.Element;
 import org.jdom.input.SAXBuilder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.uberfire.backend.server.WebAppSettings;
+import org.uberfire.backend.server.WebAppListener;
 import org.uberfire.commons.services.cdi.Startup;
 import org.uberfire.commons.services.cdi.StartupType;
 
@@ -45,6 +46,10 @@ public class RoleLoader {
     Logger logger = LoggerFactory.getLogger(RoleLoader.class);
 
     @PostConstruct
+    public void init() {
+        WebAppListener.registerOnStartupCommand(this::registerRolesFromwWebXml);
+    }
+
     public void registerRolesFromwWebXml() {
         File webXml = getWebXmlFile();
         if (webXml != null) {
@@ -64,20 +69,8 @@ public class RoleLoader {
     }
 
     protected File getWebXmlFile() {
-        String[] paths = new String[] {
-                "security-policy.properties",
-                "security-management.properties",
-                "WEB-INF/classes/security-policy.properties",
-                "WEB-INF/classes/security-management.properties"};
-
-        for (String path : paths) {
-            URL pathURL = Thread.currentThread().getContextClassLoader().getResource(path);
-            if (pathURL != null) {
-                File webInf = new File(pathURL.getPath()).getParentFile().getParentFile();
-                return new File(webInf, "web.xml");
-            }
-        }
-        return null;
+        String wefInfDir = WebAppSettings.get().getAbsolutePath("WEB-INF");
+        return new File(wefInfDir, "web.xml");
     }
 
     protected Set<String> loadRolesFromwWebXml(File webXml) throws Exception {
--- a/uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/WebAppListener.java
+++ b/uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/WebAppListener.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright 2015 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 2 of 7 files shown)


uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/security/RoleLoader.java
```diff
@@ -64,20 +69,8 @@ public void registerRolesFromwWebXml() {
...
-        String[] paths = new String[] {
-                "security-policy.properties",
-                "security-management.properties",
-                "WEB-INF/classes/security-policy.properties",
-                "WEB-INF/classes/security-management.properties"};
-
-        for (String path : paths) {
-            URL pathURL = Thread.currentThread().getContextClassLoader().getResource(path);
-            if (pathURL != null) {
-                File webInf = new File(pathURL.getPath()).getParentFile().getParentFile();
```
mswiderski: just a quick question - wouldn't it work if you use URL.inputStream instead of converting this to file?

wondering if this will work on other containers... in case we need it

dgutierr: This code no longer exists as we now rely on a ServletContenxtListener to capture the webapp's root dir.

mswiderski: true, but was wondering if use of inputStream wouldn't be simpler. anyway I guess not so important

dgutierr:
jenkins retest this

csadilek:
@dgutierr I do see a problem with this. Who/What guarantees that the servlet context is initialized after the ```PostConstruct``` method of your ```AuthorizationPolicyDeployer``` is invoked. If it runs before it there will be no callback.

csadilek:
@dgutierr to avoid this initialization ordering problem my listener gets an instance of the CDI bean and initialized it with the "web app state" (i.e. the root dir): https://github.com/csadilek/uberfire/blob/dynamic-plugins/uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/plugin/PluginServletContextListener.java

dgutierr:
As far as I know all the @Startup beans are always initialized before the any WebListener, since @Startup annotation is processed by a CDI extension.

Anyway, I think your solution is quite similar as you are also using @Startup to initialize the static RuntimePluginServiceImpl reference and the PluginServletContextListener assumes the CDI bean has been initialized. Pretty much what I'm doing, right?

csadilek:
@dgutierr no, unfortunately, you are relying on the fact that CDI initialization and ```@PostConstruct``` happens before your web listener runs. I do not, because I have been burnt by this in the past. Check you the getInstance method which will create an instance even if CDI hasn't initialized yet.

dgutierr:
@csadilek Sorry but I do not get what you mean :-). As far as I can see, the getInstance does not initialize the singleton ref:

    public static RuntimePluginServiceImpl getInstance() {
        if ( instance == null ) {
            throw new IllegalStateException( RuntimePluginService.class.getName() + " was not initialized on startup" );
        }
        return instance;
    }

Anyway, what changes do you think I should apply to get it right as you propose?

dgutierr:
I've changed the implementation. Now it does not matter what bean is initialized first. 

@csadilek Please, take a look at the WebAppListener implementation.

csadilek:
@dgutierr yes, you're right. I thought I did, but apparently I didn't in the example above from my branch. I will fix that.

+1 to your new solution which will work in either case.

ederign:
thanks @dgutierr

jhrcek:
Guys I'm not sure if this is related, but whereas in the last few day's we've been getting deployment errors related to AuthorizationPolicyDeployer initialization, I tried with today's kie-wb snapshot (timestamp 7.0.0-20160623.001847) which should contain these fixes and now I'm getting deployment failures caused by SocialTimelinePersistenceProducer.setup() [see here](https://gist.github.com/jhrcek/c8a00d24dafae3e185556e77eb876c9a). The problem is that kie-wb fails to deploy 3 times out of 4 because of this. What do you think? Should I open another jira?

dgutierr:
@csadilek I guess WebAppListener is general enough to be uised by any bean which depends on the webapp to be initialized first.  Just by adding a callback       WebAppListener.registerOnStartupCommand(Command onStartup)  any bean can be notified.

wdyt?

avkapustin:
Hello,
AuthorizationPolicyDeployer.getPolicyDir fails on windows machines with exception
```
Caused by: java.lang.IllegalArgumentException: Illegal character in authority at index 7: file://C:\dev\webui\webui-showcase\webui-webapp\target\webui-webapp\WEB-INF/classes
	at java.net.URI.create(URI.java:852)
	at org.uberfire.backend.server.authz.AuthorizationPolicyDeployer.getPolicyDir(AuthorizationPolicyDeployer.java:92)
	at org.uberfire.backend.server.authz.AuthorizationPolicyDeployer.deployPolicy(AuthorizationPolicyDeployer.java:86)
	at org.uberfire.backend.server.WebAppListener.notifyCallbacks(WebAppListener.java:87)
	at org.uberfire.backend.server.WebAppListener.setRootDir(WebAppListener.java:75)
	at org.uberfire.backend.server.WebAppListener.contextInitialized(WebAppListener.java:47)
	at io.undertow.servlet.core.ApplicationListeners.contextInitialized(ApplicationListeners.java:187)
	at io.undertow.servlet.core.DeploymentManagerImpl.deploy(DeploymentManagerImpl.java:198)
	... 8 more
Caused by: java.net.URISyntaxException: Illegal character in authority at index 7: file://C:\dev\webui\webui-showcase\webui-webapp\target\webui-webapp\WEB-INF/classes
	at java.net.URI$Parser.fail(URI.java:2848)
	at java.net.URI$Parser.parseAuthority(URI.java:3186)
	at java.net.URI$Parser.parseHierarchical(URI.java:3097)
	at java.net.URI$Parser.parse(URI.java:3053)
	at java.net.URI.<init>(URI.java:588)
	at java.net.URI.create(URI.java:850)
	... 15 more
```
Reduced test to reproduce this error
```java
public class UriTest {
    @Test(expected = IllegalArgumentException.class)
    public void testUriParseFail() {
        String path = "file://C:\\dev\\webui\\webui-showcase\\webui-webapp\\target\\webui-webapp\\WEB-INF\\classes";
        URI uri = URI.create(path);
    }

    @Test
    public void testUriParseOk() {
        String path = "file://C:/dev/webui/webui-showcase/webui-webapp/target/webui-webapp/WEB-INF/classes";
        URI uri = URI.create(path);
        Assert.assertNotNull(uri);
    }
}
```

dgutierr:
Windows issue reported as https://issues.jboss.org/browse/UF-340

It will be fixed in the next few days.

Thanks @avkapustin for the feedback!

PR #406, opened 2016-06-22
UF-328: HelpIcon - allow title/content update

mcimbora:
Using setHelpTitle/setHelpContent currently has no effect as the popover instance is not updated.

### DIFF
--- a/uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/widgets/HelpIcon.java
+++ b/uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/widgets/HelpIcon.java
@@ -30,6 +30,8 @@ public class HelpIcon extends Composite {
     private String title;
     private String content;
 
+    private Popover popover;
+
     public HelpIcon() {
         initWidget( panel );
         addStyleName( "uf-help-icon" );
@@ -39,7 +41,7 @@ public class HelpIcon extends Composite {
     protected void onLoad() {
         super.onLoad();
         if ( title != null || content != null ) {
-            final Popover popover = new Popover( icon );
+            this.popover = new Popover( icon );
             popover.setContent( content );
             popover.setTitle( title );
             popover.setContainer( "body" );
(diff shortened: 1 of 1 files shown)


uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/widgets/HelpIcon.java
```diff
@@ -39,7 +41,7 @@ public HelpIcon() {
     protected void onLoad() {
         super.onLoad();
         if ( title != null || content != null ) {
```
manstis: I know this is the original code; but it's flawed.. We should check for whether ```title != null``` then set title; and then whether ```content != null``` then set content. At the moment if title is ```null``` but content is not ```null``` we still set the title (to null).

mcimbora: @manstis Is that really a problem? Looking at Popover source https://github.com/gwtbootstrap3/gwtbootstrap3/blob/0.9.1/gwtbootstrap3/src/main/java/org/gwtbootstrap3/client/ui/Popover.java#L219 those are plain setters, so there should be no side effects.

mcimbora: i.e. The fields still remain null, as they originally were.

manstis: @mcimbora Hello, my concern was simply that setting ```title``` to null (when ```content != null```) *could* end up with the ```Popover``` title being shown as "null" (which can sometime happen, depending on the UI component's implementation).. It's hard to see what ```$wnd.jQuery(e).popover(..)``` does (in the ```Popover``` implementation) but (1) let's assume BS3 is sufficiently established to handle ```null``` values well! (2) it was only a comment on old, existing code.. and if it's not broken.. let's not fix it!

ederign: @manstis  @mcimbora are still work to do on this?

manstis: @ederign Nope, it can be +1 and merged.

manstis:
+1 to merge.

PR #407, opened 2016-06-22
UF-334: upgrade of pfly to 3.5.0

porcelli:
c3/d3 added `manually` to showcase as this is the only place that this is used (dashbuilder uses different
tech).


cristianonicolai:
+1

porcelli:
@ederign @csadilek @manstis  can you please review?

manstis:
@porcelli Looks OK to me; I'm a little uncertain about the location "where" unpacked resources end up; and the modified ```org.uberfire.UberfireShowcase/js/d3.min.js``` etc paths.. but if you've tried locally and it works OK then.... +1 from me.

ederign:
building and trying this on kie-wb

porcelli:
@csadilek ?

csadilek:
+1

PR #408, opened 2016-06-22
GUVNOR-2559: Guided Decision Table Editor: Improve appearance (for Red Hat Summit) - Some colour changes and add icon to Linked columns

manstis:
Changed some mouse handlers to only try alternative checks if a match was not found by the prior check. There was a problem that selecting a "linked" column (which selects the "linked" grid) also was selecting the grid that contained the column clicked on - because a single "click" was handled twice; once to select the linked grid, and once to select the grid clicked on. Unit Test changes reflect the change in behaviour.


csadilek:
+1

manstis:
@ederign @csadilek @porcelli Could you please merge this when you get a chance. Thanks.

PR #414, opened 2016-06-27
upgraded errai to 3.2.4.Final

mbiarnes:


psiroky:
+1, we need to fix those randomly failing tests (or remove them)!

PR #421, opened 2016-06-27
UF-335: Fix NPE when injected SessionInfo used outside of request, an…

mbarkley:
…d check for non-null filesystems in AbstractIOService.

This is a partial for UF-335. See the [issue](https://issues.jboss.org/browse/UF-335) comments for details.

### DIFF
--- a/uberfire-server/src/main/java/org/uberfire/server/cdi/UberFireGeneralFactory.java
+++ b/uberfire-server/src/main/java/org/uberfire/server/cdi/UberFireGeneralFactory.java
@@ -16,11 +16,13 @@
 
 package org.uberfire.server.cdi;
 
+import static org.jboss.errai.bus.server.api.RpcContext.getMessage;
+import static org.jboss.errai.bus.server.api.RpcContext.getQueueSession;
+
 import javax.enterprise.context.RequestScoped;
 import javax.enterprise.inject.Default;
 import javax.enterprise.inject.Produces;
 
-import org.jboss.errai.bus.server.api.RpcContext;
 import org.jboss.errai.security.shared.service.AuthenticationService;
 import org.uberfire.rpc.SessionInfo;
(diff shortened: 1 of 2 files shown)


uberfire-server/src/main/java/org/uberfire/server/cdi/UberFireGeneralFactory.java
```diff
@@ -31,7 +33,11 @@
     @RequestScoped
     @Default
     public static SessionInfo getSessionInfo(AuthenticationService authenticationService) {
-        return new SessionInfoImpl( RpcContext.getQueueSession().getSessionId(), authenticationService.getUser() );
+        return new SessionInfoImpl( getSessionInfo(), authenticationService.getUser() );
+    }
+
+    private static String getSessionInfo() {
+        return (getMessage() != null && getQueueSession() != null ? getQueueSession().getSessionId() : null );
```
manstis: @mbarkley Out of interest what's the significance of ```getMessage()``` it's not used explicitly in ```getSessionInfo()```?

manstis:
+1 to merge.

csadilek:
+1

ederign:
thanks @mbarkley

uberfire-server/src/main/java/org/uberfire/server/cdi/UberFireGeneralFactory.java
mbarkley: The queue session is stored in the message returned by `getMessage()`.

jhrcek: So it means there's hidden potential for NPE in every call to `getQueueSession()` when `getMessage()` returns null? Shouldn't that be handled in `getQueueSession()` directly?

PR #422, opened 2016-06-28
GUVNOR-2556: Right-clicking a cell should select it

manstis:
See https://issues.jboss.org/browse/GUVNOR-2556

Moved the logic relating to the identification of a i"selection" into a utility class and wrote some tests.


manstis:
See https://github.com/droolsjbpm/drools-wb/pull/225 too.

csadilek:
+1

PR #423, opened 2016-06-28
Return default user outside of HTTP session threads.

mbarkley:
This fixes issues described in [UF-335](https://jira.jboss.org/browse/UF-335).


csadilek:
+1

mbarkley:
After working with Christian, we have fixed the SessionInfo produced by the UberFireGeneralFactory so that it is safe to access outside of HTTP session threads. We have been able to successfully deploy the kie-wb-webapp with this fix with org.kie.example=true.

PR #429, opened 2016-07-01
GUVNOR-2597: Guided Decision Table Editor: Exiting 'pinned mode' doesn't restore correct TransformMediator

manstis:
See https://issues.jboss.org/browse/GUVNOR-2597


manstis:
See https://github.com/droolsjbpm/drools-wb/pull/230 too.

Rikkola:
+1

manstis:
@ederign @csadilek Could you please merge?

PR #433, opened 2016-07-08
Security management enhancements

dgutierr:
ACL Editor CSS polishment
UF-333: Stored values are not correctly presented in Security perspective settings


dgutierr:
@ederign Can you verify please?

ederign:
+1 thanks @dgutierr

uberfire-backend/uberfire-backend-server/src/test/java/org/uberfire/backend/server/authz/AuthzPolicyMarshallerTest.java
```diff
@@ -80,6 +80,16 @@ public void testPermissionEntry() {
         assertEquals(tokens.get(3), "perspective.read");
     }
 
+    @Test
+    public void testPermissionEntry2() {
```
csadilek: @ederign @dgutierr please always use descriptive test names! The test name should reflect what the test is about. We don't want test1, testX2, etc.

PR #434, opened 2016-07-08
Upgrading component archetype for Errai UI and also removing old (unmaintained) project archetype

ederign:
…


ederign:
I've also found another issue to be address in the future: https://issues.jboss.org/browse/UF-358

paulovmr:
+1 to merge.

PR #435, opened 2016-07-09
Syntax improvement on the documentation

danilopereira:
In order to look through the context of this guide, I think it is the better way to say it.


ederign:
Hi @danilopereira first of all, thanks for your contribution. However, the current version makes more sense in this context, so I'll keep the current version.

We really appreciate new contributions to the Uberfire, and the tutorial is really the best place to start.

Thanks!

PR #436, opened 2016-07-11
[UF-357] put jboss repos in the archetype pom

ederign:


ederign:
@psiroky could you please take a look on this?

psiroky:
+1, looks good.

Nice to have: We could add a comment explaining why the repo is needed (as users who generate the archetype may wonder about that).

PR #438, opened 2016-07-12
UF-359: Client-only showcase/demo does not compile and is outdated

csadilek:
This fixes the client-only showcase which wasn't being compiled. Once the GWT compiler was activated it actually failed to compile.

There were some minor other issues which are fixed here as well.

@ederign please ;). This demo is really useful as it shows how to run UberFire without a server. Something we're going to do in our Eclipse integration effort but I believe is generally useful to the community. If we find the time we should improve this demo further so it better showcases UberFire and isn't just a hello world type demo.


manstis:
+1 seems innocuous enough to me (although Shadow Services sound intriguing!)

uberfire-showcase/uberfire-client-webapp/src/main/webapp/index.html
```diff
@@ -20,26 +20,39 @@
...
+            </div>
+        </div>
     </div>
 </div>
 
 <!-- The GWT js file generated at run time -->
 <script type="text/javascript" src='org.uberfire.UberfireShowcaseClient/org.uberfire.UberfireShowcaseClient.nocache.js'></script>
 
-<script type="text/javascript">
-    erraiBusRemoteCommunicationEnabled = false;
```
porcelli: nice we don't need this anymore ;)

porcelli:
+1

ederign:
@csadilek great work! Just tested in Apache: 
screen shot 2016-07-12 at 3 04 36 pm

PR #441, opened 2016-07-14
This is a OS' prerequisite

danilopereira:
This is in a comment in the pom.xml, but we need it to compile the project, so it is a prerequisite


ederign:
Hi @danilopereira, this is only a requirement for maven-graph-plugin and not for the Uberfire.

PR #442, opened 2016-07-14
UF-305 - PatternFly upgrade to version 3.6.0

cristianonicolai:
Style fixes for upgrade
New theme GWT module

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -62,7 +62,7 @@
     <version.javax.validation>1.0.0.GA</version.javax.validation>
     <version.org.jboss.errai>4.0.0-SNAPSHOT</version.org.jboss.errai>
     <version.org.jboss.keycloak>1.9.0.Final</version.org.jboss.keycloak>
-    <version.org.webjars.bower.org.patternfly>3.5.0</version.org.webjars.bower.org.patternfly>
+    <version.org.webjars.bower.org.patternfly>3.6.0</version.org.webjars.bower.org.patternfly>
     <version.org.webjars.bower.google-code-prettify>1.0.4</version.org.webjars.bower.google-code-prettify>
 
     <version.org.picketlink>2.6.0.Final</version.org.picketlink>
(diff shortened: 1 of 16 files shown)


cristianonicolai:
@manstis @csadilek changes I mentioned regarding PatternFly, please have a look.
Please note that showcase apps need to be updated in the different repos to use the new GWT module: org.uberfire.client.views.PatternFlyTheme

ederign:
Hi @cristianonicolai ! 

Why move PatternFly -> PatternFlyTheme ? And also, why keep both?

@csadilek @cristianonicolai @manstis do you guys have any plan in order to updgrade other showcases?

csadilek:
+1

@ederign that was done to support different themes. The default community theme (PatternFlyBase->PatternFlyTheme) and a theme for the product (PatternFlyBase->KIETheme, based on rcue).

But, I am not sure why there are 3 modules ;) (PatternFlyBase, PatternFly and PatternFlyTheme).

@cristianonicolai can you clarify why there's an additional PatternFly.gwt.xml. Wouldn't it be enough to have PatternFlyBase->PatternFlyTheme and PatternFlyBase->KIETheme. In other words, why is KIETheme inheriting PatternFly which is then inheriting PatternFlyBase? Couldn't KIETheme directly inherit PatternFlyBase (like PatternFlyTheme does), esp. as PatternFly doesn't seem to add anything?

csadilek:
@cristianonicolai to clarify would it be enough to have PatternFlyBase and PatternFlyTheme in UberFire?

Then have: 
KIETheme (product) inherit from PatternFlyBase, + rcue styles
KIETheme (community) inherit from PatternFlyTheme

cristianonicolai:
@csadilek @ederign I agree, we have an extra module that we could remove from here. PatternFly module needs to stay basically to keep compatibility with all the GWT modules that inherits from it. Otherwise we would have to fix all of those just to rename from PatternFly to PatternFlyBase.
That way we would have:
KIETheme (product) inherit from PatternFly, + rcue styles
KIETheme (community) inherit from PatternFlyTheme
PatternFlyTheme inherit from PatternFly + patternfly styles
WDYT?

csadilek:
@cristianonicolai ok, got it, thanks. I am fine either way.

cristianonicolai:
@ederign I've pushed PRs for most webapps/showcases that I could find, hopefully I haven't missed any.

PR #446, opened 2016-07-22
Fix the CopyPopUpView injection

karreiro:
Creates a new qualifier `@DefaultView` that solves the CDI ambiguity.

### DIFF
--- a/uberfire-extensions/uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/file/CopyPopUpViewProducer.java
+++ b/uberfire-extensions/uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/file/CopyPopUpViewProducer.java
@@ -22,6 +22,7 @@ import javax.enterprise.inject.Produces;
 import javax.inject.Inject;
 import javax.inject.Named;
 
+import org.jboss.errai.ioc.client.api.ManagedInstance;
 import org.uberfire.commons.services.cdi.Startup;
 import org.uberfire.commons.services.cdi.StartupType;
 import org.uberfire.ext.editor.commons.client.file.popups.CopyPopUpPresenter;
@@ -34,11 +35,15 @@ public class CopyPopUpViewProducer {
     @Inject
     private Instance<CopyPopUpPresenter.View> copyPopUpViewInstance;
 
+    @Inject
+    @DefaultView
+    private ManagedInstance<CopyPopUpView> defaultCopyPopUpViewProvider;
+
     @Produces
     @Customizable
     public CopyPopUpPresenter.View copyPopUpViewProducer() {
         if ( this.copyPopUpViewInstance.isUnsatisfied() ) {
-            return new CopyPopUpView();
+            return defaultCopyPopUpViewProvider.get();
         }
 
         return this.copyPopUpViewInstance.get();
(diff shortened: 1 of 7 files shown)


paulovmr:
+1 to merge.

ederign:
I'll try to run it locally and then merge.

paulovmr:
+1 to merge.

manstis:
@csadilek @ederign We really need this merged soon (all workbench webapps are currently broken)

karreiro:
@ederign I've just fixed the alignment issue that you saw.

ederign:
@manstis  I'm just waiting to "green" from jenkins.

ederign:
thanks @karreiro

manstis:
Thanks everyone :-)

mbarkley:
I see that this is merged already, but just thought I'd give some feedback in case anyone has time to clean up or enhance this code in the future:

* I think `@FallbackView` would be a better name than `@DefaultView` (because really it's the implementation you fallback on when there is no `@Default`).

* Similarly `defaultCopyPopUpViewProvider` would be better named `fallbackCopyPopUpViewProvider`.

* It looks odd now that `copyPopUpViewInstance` is an `Instance` while `defaultCopyPopUpViewProvider` is a `ManagedInstance`. They should probably both be `ManagedInstance`.

manstis:
Max, do we need ManagedInstance? Both instances are created through CDI interfaces so won't the underlying BeanManager handle clean up already? Fallback vs Default is, in my opinion, a matter of personal preference... personally a qualifier to specify an "override" is more clear.. but I'm conscious my opinions are just that: opinions and the way somebody else implements something need not be wrong just different. This sounds like a rant... it's not; so apologies if it reads that way.

manstis:
Fallback implies, to me, there's an expectation you *should* provide your own implementation otherwise it'll fallback to an erm.. default ;-) Default suggests to me you *could* provide your own implementation.. but this should be discussed over a couple if beers :-)

mbarkley:
Hey Michael,

Yeah, naming is subjective but I have a less subjective concern I should have been more direct about: both implementations have qualifiers with the word "default" in them (because one is `@DefaultView` and the other is implicitly `@Default`). With that in mind, I suggested "fallback" as a remedy but there are any number of other names not involving the word "default" that I would like equally. But if this doesn't bother or confuse others then I can live with that.

Regarding the `ManagedInstance` it should almost always be preferred over a plain `Instance`. Using `Instance` or the bean manager directly provides no automatic clean up; you are taking responsibility for manually managing the life-cycle of any `@Dependent` beans you instantiate. So `Instance` is only as good as `ManagedInstance` if you are certain that every bean you instantiate is `@ApplicationScoped`/`@Singleton`.

That being said there is still a potential memory leak (if the beans created by the `Instance`/`ManagedInstance` are `@Dependent`) with the producer in this commit that isn't fixed simply by using `ManagedInstance`. Assuming the produced beans are dependent, one of these conditions needs to be met for them to be cleaned up:

1. `destroyBean` is called on the produced instances.
2. The bean instance is created by a `ManagedInstance` which is destroyed.

The reason (2) doesn't happen in this case is because the producer type is `@ApplicationScoped`, so the field-injected `ManagedInstance` will be in service as long as the app is running.

Some solutions that fix this:

* Use only `ManagedInstance` and make the producer type `@Dependent`.
* Use only `ManagedInstance` but instead of injecting them as fields, add them as parameters of the producer method.
* Add a `@Disposes` method in the producer type that calls `destroyBean`.

The first two solutions are preferable, but the third can be appropriate if there is other clean-up code that should be executed (so probably not in this case).

manstis:
@mbarkley Thanks for your detail reply :-)

I see your point re: ```@DefaultView``` vs ```@Default``` (which would be the match for ```copyPopUpViewInstance```). It is a bit confusing.. The same clarification can be obtained if we moved the qualifiers from the "default" to the "custom" i.e. ```copyPopUpViewInstance``` is qualified with (for example) ```@CustomCopyPopup``` and the implementation in ```kie-wb-common``` uses the same qualifier. Use of ```@DefaultView``` removed.

@karreiro Could you please consider this discussion; and implement which ever you prefer (as long as we move away from ```@DefaultView``` and ```@Default```). Also please consider @mbarkley second option to ensure ```CopyPopUpView``` instances are destroyed.

karreiro:
Ok @manstis and thanks @mbarkley :-)

PR #449, opened 2016-07-25
UF-280: Security management README tweaks

dgutierr:
@ederign A README file I forgot to update during UF-280 development.


ederign:
+1 thanks @dgutierr

PR #450, opened 2016-07-26
UF-319: Workbench preferences screen

paulovmr:


uberfire-bom/pom.xml
```diff
@@ -614,6 +614,19 @@
...
+        <version>${project.version}</version>
+      </dependency>
+
+      <dependency>
+        <groupId>org.uberfire</groupId>
+        <artifactId>uberfire-preferences-processors</artifactId>
+        <version>${project.version}</version>
+        <classifier>sources</classifier>
+      </dependency>
+
```
csadilek: Is it really worth it to create a new maven artifact? Couldn't this be part of uberfire-workbench-processors? Don't feel strongly about it but wondering if there's a reason for it?

uberfire-extensions/uberfire-preferences/uberfire-preferences-api/src/main/java/org/uberfire/ext/preferences/shared/PreferenceScope.java
```diff
@@ -31,4 +31,10 @@
      * @return Key of the preference scope. Represents a unique key inside a scope type.
      */
     String key();
+
+    /**
+     * Child scope, an scope inside an scope. This allows a hierarchy inside each scope.
+     * @return Child scope. This can be null, indicating the end of the hierarchy.
```
csadilek: a(n) scope.

uberfire-bom/pom.xml
paulovmr: Hi @csadilek , the reason is that all preferences code are in uberfire-extensions, so wouldn't make sense to keep the processors in uberfire-workbench-processors (unless we move all of it to uberfire itself).

uberfire-extensions/uberfire-preferences/uberfire-preferences-api/src/main/java/org/uberfire/ext/preferences/shared/PreferenceScopeFactory.java
```diff
@@ -0,0 +1,85 @@
...
+     * @param type Type of the new preference scope.
+     * @return A new preference scope.
+     * @throws InvalidPreferenceScopeException if the type passed is invalid, or if it requires a key.
+     */
+    PreferenceScope createScope( String type,
+                                 PreferenceScope childScope ) throws InvalidPreferenceScopeException;
+
+    /**
+     * Builds a new preference scope, that requires a custom key.
+     * If the type passed does not require a custom key, a exception is thrown.
```
csadilek: Don't really need to put it here as you mention it under ```@throws```.

uberfire-extensions/uberfire-preferences/uberfire-preferences-api/src/main/java/org/uberfire/ext/preferences/shared/impl/DefaultPreferenceScopeTypes.java
```diff
@@ -14,20 +14,26 @@
...
 import org.uberfire.ext.preferences.shared.impl.exception.InvalidPreferenceScopeException;
 import org.uberfire.rpc.SessionInfo;
 
-@Vetoed
+/**
+ * Default implementation for {@link PreferenceScopeTypes}. To override it, just provide a default
+ * CDI bean that implements {@link PreferenceScopeTypes}.
+ */
+@ApplicationScoped
+@FallbackImplementation
```
csadilek: I don't see why we need FallbackImplementation in this case. The javaDoc should read...to override it just provide a CDI bean that implements PreferencesScopeType and annotate it with ```@Customizable```. It seems having both Customizable and FallbackImplementation isn't needed as your application scoped producer creates the default implementation, if unsatisfied, or the customizable one if provided. So you can use the same mechanism you have for the resolution strategy, no?

uberfire-extensions/uberfire-preferences/uberfire-preferences-api/src/main/java/org/uberfire/ext/preferences/shared/impl/PreferenceScopeImpl.java
```diff
@@ -60,14 +70,18 @@ public boolean equals( final Object o ) {
...
+        return !( childScope != null ? !childScope.equals( that.childScope ) : that.childScope != null );
 
     }
 
     @Override
     public int hashCode() {
         int result = type != null ? type.hashCode() : 0;
         result = 31 * result + ( key != null ? key.hashCode() : 0 );
+        result = 31 * result + ( childScope != null ? childScope.hashCode() : 0 );
         return result;
```
csadilek: For hashCode in portable classes (classes that are potentially used both on the client and the server) we have to make sure they behave the same in JavaScript and Java. In JavaScript an integer is between -2^53-1 and 2^53-1, whereas in Java integers are between -2^31 and 2^31-1. See this recent commit for a workaround how to achieve that: https://github.com/droolsjbpm/jbpm-console-ng/commit/26733de19bb6dc0ac84ea9b140deaf77e87709e3

While “~~” doesn’t do anything in Java (double bit-wise negation), it coerces a number into 32 bit in JavaScript.

uberfire-extensions/uberfire-preferences/uberfire-preferences-client/src/main/java/org/uberfire/ext/preferences/client/ioc/scope/PreferenceScopeResolutionStrategyIOCProvider.java
```diff
@@ -0,0 +1,55 @@
...
+import org.jboss.errai.ioc.client.api.ManagedInstance;
+import org.uberfire.ext.preferences.client.ioc.annotations.ComponentKey;
+import org.uberfire.ext.preferences.client.ioc.store.PreferenceStore;
+import org.uberfire.ext.preferences.client.ioc.utils.PreferenceQualifierUtils;
+import org.uberfire.ext.preferences.shared.PreferenceScopeFactory;
+import org.uberfire.ext.preferences.shared.PreferenceScopeResolutionStrategy;
+import org.uberfire.ext.preferences.shared.annotations.Customizable;
+import org.uberfire.ext.preferences.shared.impl.DefaultPreferenceScopeResolutionStrategy;
+
+@IOCProvider
```
csadilek: Why doe we use an IOC provider here instead of a CDI producer like in the other cases?

uberfire-bom/pom.xml
csadilek: @paulovmr ah ok, makes sense. Didn't expect preferences to be an extension. Seems worthy of being a core feature.

csadilek:
@paulovmr nice work! A couple of comments/questions inline. Make sure the get UXD to take a quick look also, if you haven't already done so. Thanks.

uberfire-extensions/uberfire-preferences/uberfire-preferences-api/src/main/java/org/uberfire/ext/preferences/shared/impl/DefaultPreferenceScopeTypes.java
paulovmr: @csadilek Actually, to override the fallback implementation, a CDI bean annotated (probably implicitly) with @Default should be provided. That is why there is a need for the @FallbackImplementation qualifier, to prevent that this bean be injected by default. This was achieved with @Vetoed before, but turns out that @Vetoed doesn't work on the client-side.

uberfire-extensions/uberfire-preferences/uberfire-preferences-client/src/main/java/org/uberfire/ext/preferences/client/ioc/scope/PreferenceScopeResolutionStrategyIOCProvider.java
paulovmr: @csadilek Because I can't have an InjectionPoint instance as a parameter for the producer method on the client-side, and since I had to look at the injection qualifiers, the IOC provider was the only alternative that I found.

uberfire-extensions/uberfire-preferences/uberfire-preferences-api/src/main/java/org/uberfire/ext/preferences/shared/impl/DefaultPreferenceScopeTypes.java
manstis: @karreiro @mbarkley @paulovmr @csadilek Would be an idea to move ```@FallbackImplementation``` to somewhere more "shareable" since Max proposed such an annotation relating to recent changes for [popups](https://github.com/uberfire/uberfire/pull/448)

uberfire-bom/pom.xml
paulovmr: @csadilek Oh I see. Do you think we should discuss moving it or leave it where it is?

csadilek: @paulovmr fine for now, we can move it later, imo.

uberfire-extensions/uberfire-preferences/uberfire-preferences-api/src/main/java/org/uberfire/ext/preferences/shared/impl/DefaultPreferenceScopeTypes.java
paulovmr: @manstis I agree, I'll move it to uberfire-api, unless anyone objects.

manstis: @paulovmr No objection here.. can be a later PR if you like.. could you also work with @karreiro to refactor the popup's ```@DefaultView``` when time permits?

uberfire-extensions/uberfire-preferences/uberfire-preferences-client/src/main/java/org/uberfire/ext/preferences/client/ioc/scope/PreferenceScopeResolutionStrategyIOCProvider.java
mbarkley: The way this provider is currently, the instances produced by `scopeResolutionStrategyProvider` will never be destroyed. One of the following would fix this:

* Change the provider to `@Dependent`.
* Implement `Disposer` like [this provider](https://github.com/errai/errai/blob/master/errai-ioc/src/main/java/org/jboss/errai/ioc/client/api/builtin/ManagedInstanceProvider.java#L45) and call destroyBean manually.

uberfire-extensions/uberfire-preferences/uberfire-preferences-api/src/main/java/org/uberfire/ext/preferences/shared/impl/DefaultPreferenceScopeTypes.java
csadilek: @paulovmr I don't get it. You never need to be able to override the fallback version (it's the default in case a user didn't provide their own).  If a user does provide their own it will be satisfied and your producer will return that customized instance instead, otherwise the producer will new() it (as you do for resolution strategy). So, your PreferenceScopeTypes shouldn't be ```@ApplicationScoped``` so it's unsatisfied by default and be newed int he producer.

uberfire-extensions/uberfire-preferences/uberfire-preferences-client/src/main/java/org/uberfire/ext/preferences/client/ioc/scope/PreferenceScopeResolutionStrategyIOCProvider.java
csadilek: @paulovmr @mbarkley ah that would be a good JIRA/feature request for Errai to support producers that get passed an injection point. That's supported on the server.

mbarkley: @csadilek @paulovmr I looked into this a while ago and the only issue is the InjectionPoint API uses lots of `java.lang.reflect.*` classes and CDI classes. I'm not sure it's worth the effort to emulate that API, but maybe we could have a ClientInjectionPoint with a similar but client-specific API?

csadilek: @mbarkley @paulovmr sounds good to me to only support a subset of InjectionPoint functionality on the client. It's not a high priority feature, but nice-to-have as you'd expect it work when looking at the CDI spec.

uberfire-extensions/uberfire-preferences/uberfire-preferences-api/src/main/java/org/uberfire/ext/preferences/shared/impl/DefaultPreferenceScopeTypes.java
paulovmr: @manstis Sure!

@csadilek Ok, I think I expressed myself in the wrong way: by override I mean provide a custom implementation. Let's see if I understood correctly: You are proposing that `DefaultPreferenceScopeTypes.java` be annotated with nothing, and then its producer new() it to provide an instance when necessary? If that is what you meant, the problem is that it'll still be recognized as a default bean by CDI (`@Dependent`, since no scope was used), and if the user provide its own default bean, there will happen an ambiguous resolution error. If that's not what you meant, sorry, I didn't get it.

uberfire-extensions/uberfire-preferences/uberfire-preferences-client/src/main/java/org/uberfire/ext/preferences/client/ioc/scope/PreferenceScopeResolutionStrategyIOCProvider.java
paulovmr: @mbarkley Thanks! I will change that.

@csadilek @mbarkley I agree with the subset support.

paulovmr:
@csadilek Thanks for the review! UXD is working on a screen for the KIE Workbench, which development is the next task related to preferences. The screens that are on this PR are inside uberfire-wires-webapp, and only serves test/demonstration purposes. They will probably be removed after the UXD screen is developed.

uberfire-extensions/uberfire-preferences/uberfire-preferences-api/src/main/java/org/uberfire/ext/preferences/shared/impl/DefaultPreferenceScopeTypes.java
csadilek: @paulovmr let's chat about it, maybe I explained badly or got confused.

paulovmr:
@csadilek The changes are done. Thanks for the review!

General improvements: https://gist.github.com/paulovmr/03b9b093851e5eb463fed2e294327a61
Moving @Customizable and @FallbackImplementation to uberfire-api: https://gist.github.com/paulovmr/790949c6e3eea7f1cf4762bdcb23e85b

csadilek:
+1

ederign:
thanks @paulovmr great work!

We just have to make sure that this temporary screen in wires doesn't go to product someday.

PR #454, opened 2016-07-28
GUVNOR-2624: Guided Decision Table Editor: Selection range does not extend correctly with keyboard navigation

manstis:
See https://issues.jboss.org/browse/GUVNOR-2624


ederign:
+1 thanks @manstis

PR #461, opened 2016-08-05
Upgrade to GWT 2.8.0-rc1

csadilek:
Most external dependencies that used to be inside the gwt-dev.jar have now been unbundled. So we can get rid of the enforcer exception and can now manage exclusions using the standard maven mechanism.


csadilek:
Needs:

https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/250
https://github.com/errai/errai/pull/184

uberfire-extensions/uberfire-security/uberfire-security-management/pom.xml
```diff
@@ -42,5 +42,56 @@
     <module>uberfire-security-management-webapp</module>
   </modules>
 
+  <dependencyManagement>
```
manstis: Shouldn't this be in https://github.com/uberfire/uberfire/blob/master/pom.xml

csadilek: @manstis no it only affects the security modules. the relevant exclusions for everything else are in uf master.

manstis:
+1 to merge (but ```ecj``` comment probably needs investigating as suggested by @psiroky)

uberfire-extensions/uberfire-security/uberfire-security-management/pom.xml
psiroky: We should generally _not_ use `dependencyManagement` in submodules as it creates lots of different issues.

Having said, its very probably fine in this case since these deps won't be used at runtime (e.g. in case the exclusion was on some library which is then bundled in WAR, we would end-up in situation where the security modules are tested _without_ this dependency as it is excluded, but in fact at runtime, the dependency would be there as the WAR may not exclude it).

csadilek: @psiroky right these are all for provided deps. Also can't exclude them on master as then I would have to add a replacement in for other modules. This seemed unnecessary to me...

csadilek:
jenkins retest this.

PR #466, opened 2016-08-17
ObjectStorage implementation

adrielparedes:
This is an abstraction of a Git Storage, it is based on @paulovmr PreferenceStorage.

@ederign 
@porcelli

### DIFF
--- a/uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/io/object/ObjectStorage.java
+++ b/uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/io/object/ObjectStorage.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.uberfire.backend.server.io.object;
+
+import org.uberfire.java.nio.file.Path;
+
+/**
+ * Object Storage definition
+ */
+public interface ObjectStorage {
+
+    /**
+     * Initialize the object storage.
+     * @param rootPath The root path where is going to create the file system
+     */
+    void init( String rootPath );
+
+    /**
+     * Checks if a path exists into the object storage.
+     * @param path the path to the stored object.
+     * @return true if exists that path into the storage else false.
+     */
+    boolean exists( String path );
+
+    /**
+     * Read the content of the path given
+     * @param path The path where the content is stored
+     * @return The content of the file unmarshalled
+     */
+    <T> T read( String path );
+
+    /**
+     * Write an object into the object storage. That object is marshelled into JSON.
+     * @param path The path where the object is going to be written
+     * @param value The object itself
+     */
+    <T> void write( String path,
+                    T value );
+
+    /**
+     * Delete a path from the object storage if exists. That path contains the object stored.
+     * @param path The path to delete
+     */
+    void delete( String path );
+
+    /**
+     * Return a path based on the filesystem that is implemented for storage
+     * @param first the path string or initial part of the path string
+     * @param path additional strings to be joined to form the path string
+     * @return the path object that represents the path.
+     */
+    Path getPath( String first,
+                  String... path );
+}
--- a/uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/io/object/ObjectStorageImpl.java
+++ b/uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/io/object/ObjectStorageImpl.java
@@ -1,0 +1,110 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.uberfire.backend.server.io.object;
+
+import java.net.URI;
+import java.util.HashMap;
+import javax.inject.Inject;
+import javax.inject.Named;
+
+import org.jboss.errai.marshalling.server.ServerMarshalling;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.uberfire.io.IOService;
+import org.uberfire.java.nio.file.FileSystem;
+import org.uberfire.java.nio.file.FileSystemAlreadyExistsException;
+import org.uberfire.java.nio.file.Path;
+
+public class ObjectStorageImpl implements ObjectStorage {
+
+    private static final Logger logger = LoggerFactory.getLogger( ObjectStorageImpl.class );
+
+    private IOService ioService;
+
+    private FileSystem fileSystem;
+
+    @Inject
+    public ObjectStorageImpl( @Named("configIO") final IOService ioService ) {
+        this.ioService = ioService;
+    }
+
+    @Override
+    public void init( String rootPath ) {
+        try {
+            fileSystem = ioService.newFileSystem( URI.create( rootPath ),
+                                                  new HashMap<String, Object>() {{
+                                                      put( "init", Boolean.TRUE );
+                                                      put( "internal", Boolean.TRUE );
+                                                  }} );
+        } catch ( FileSystemAlreadyExistsException e ) {
+            fileSystem = ioService.getFileSystem( URI.create( rootPath ) );
+        }
+    }
+
+    @Override
+    public boolean exists( final String path ) {
+        Path fsPath = fileSystem.getPath( path );
+
+        try {
+            return ioService.exists( fsPath );
+        } catch ( final Exception e ) {
+            throw new RuntimeException( e );
+        }
+    }
+
+    @Override
+    public <T> T read( final String path ) {
+        Path fsPath = fileSystem.getPath( path );
+        try {
+            if ( ioService.exists( fsPath ) ) {
+                String content = ioService.readAllString( fsPath );
+                return (T) ServerMarshalling.fromJSON( content );
+            }
+        } catch ( final Exception e ) {
+            throw new RuntimeException( e );
+        }
+
+        return null;
+    }
+
+    @Override
+    public <T> void write( final String path,
+                           final T value ) {
+        try {
+            ioService.startBatch( fileSystem );
+            Path fsPath = fileSystem.getPath( path );
+            String content = ServerMarshalling.toJSON( value );
+            ioService.write( fsPath, content );
+        } catch ( final Exception e ) {
+            throw new RuntimeException( e );
+        } finally {
+            ioService.endBatch();
+        }
+
+    }
+
+    @Override
+    public void delete( final String path ) {
+        ioService.deleteIfExists( fileSystem.getPath( path ) );
+    }
+
+    @Override
+    public Path getPath( String first,
+                         String... paths ) {
+        return this.fileSystem.getPath( first, paths );
+    }
+}
--- a/uberfire-backend/uberfire-backend-server/src/test/java/org/uberfire/backend/server/io/object/ObjectStorageTest.java
+++ b/uberfire-backend/uberfire-backend-server/src/test/java/org/uberfire/backend/server/io/object/ObjectStorageTest.java
@@ -1,0 +1,156 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.uberfire.backend.server.io.object;
+
+import java.io.File;
+import java.io.IOException;
+import java.net.URI;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.jboss.errai.marshalling.client.MarshallingSessionProviderFactory;
+import org.jboss.errai.marshalling.client.protocols.MarshallingSessionProvider;
+import org.jboss.errai.marshalling.server.MappingContextSingleton;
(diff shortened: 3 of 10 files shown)


uberfire-backend/uberfire-backend-server/src/test/java/org/uberfire/backend/server/io/object/ObjectStorageTest.java
```diff
@@ -0,0 +1,144 @@
...
+        objectStorage = new ObjectStorageImpl( ioService, "/object/store" );
+    }
+
+    @After
+    public void cleanupFileSystem() {
+        fileSystemTestingUtils.cleanup();
+    }
+
+    @Test
+    public void testForAKeyDoesNotExist() {
```
paulovmr: The "key" concept doesn't make sense in the ObjectStorage. I think "file" is a better name.

uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/io/object/ObjectStorage.java
```diff
@@ -0,0 +1,64 @@
...
+     */
+    void delete( String path );
+
+    /**
+     * Return a path based on the filesystem that is implemented for storage
+     * @param first
+     * @param path
+     * @return
+     */
+    Path getPath( String first,
```
paulovmr: I think there could be more details in this javadoc about the parameters. It isn't clear what each one represents.

uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/io/object/ObjectStorageImpl.java
```diff
@@ -0,0 +1,115 @@
...
+        return null;
+    }
+
+    @Override
+    public void write( final String path,
+                       final Object value ) {
+        try {
+            ioService.startBatch( fileSystem );
+            Path fsPath = fileSystem.getPath( path );
+            ioService.write( fsPath, xs.toXML( value ) );
```
mrietveld: Apologies if I'm missing something, but using something like Kryo seems like a better idea than using XStream, given that Xstream is more limited in it's serialization capabilities than something like Kryo. 

(Or is there a guarantee here that "Object" will always be an Xstream serializable object?)

porcelli: @mrietveld you're correct; we've beeing using for a while in UF/Workbench ecosystem to store configs. btw: Objects are configs, and vice versa.

@ederign what you think about get rid of xstream and adopt Kryo, and for marshalling to/from human readable, I'd go for json.

uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/io/object/ObjectStorageImpl.java
```diff
@@ -0,0 +1,115 @@
...
+
+    private IOService ioService;
+
+    private FileSystem fileSystem;
+
+    private final XStream xs = new XStream();
+
+    public ObjectStorageImpl( final IOService ioService,
+                              String rootPath ) {
+        this.ioService = ioService;
```
ederign: why not pass a FileSystem instance for this?

uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/io/object/ObjectStorageImpl.java
ederign: @porcelli I would also go for Json + Gson but seems like @paulovmr doesn't like my idea. :(

uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/io/object/ObjectStorageImpl.java
ederign: I mean, for me it's to much responsibility for ObjectStorage to create a filesystem if there isn't one.

porcelli: the whole point of this abstraction is to hide storage details.. so -1 to even pass IOService, this should be hidden for external.

uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/io/object/ObjectStorageImpl.java
paulovmr: @ederign My only restriction (maybe a obvious one) is that the deserialization must restore the same instance for each attribute object. If Kryo and Gson do that, I'm happy with either one. :)

mrietveld: +1 on using JSON instead of Xstream/XML -- in my experience with the REST services, JSON is far more permissive and fail-friendly than XML.

ederign: You can use Type Adapters for Gson.

uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/io/object/ObjectStorageImpl.java
paulovmr: @adrielparedes I agree with @porcelli , I think we can go back to that idea of injecting the IOService in the constructor, annotate the class with `@Dependent` and pass the `rootPath` through an initializer method, which will create the file system.

porcelli: @Dependent is default behavior

uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/io/object/ObjectStorageImpl.java
```diff
@@ -0,0 +1,115 @@
...
+import org.uberfire.java.nio.file.FileSystemAlreadyExistsException;
+import org.uberfire.java.nio.file.FileVisitResult;
+import org.uberfire.java.nio.file.Path;
+import org.uberfire.java.nio.file.SimpleFileVisitor;
+import org.uberfire.java.nio.file.attribute.BasicFileAttributes;
+
+import static org.uberfire.commons.validation.PortablePreconditions.*;
+import static org.uberfire.java.nio.file.Files.*;
+
+public class ObjectStorageImpl implements ObjectStorage {
```
ederign: Also why this is not a CDI bean?

uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/io/object/ObjectStorageImpl.java
adrielparedes: Is there any way to get the IOService without injection? I mean, If we want to use the ObjectStorage in a low level API (JGit for instance).

uberfire-backend/uberfire-backend-server/src/test/java/org/uberfire/backend/server/io/object/ObjectStorageTest.java
```diff
@@ -0,0 +1,144 @@
...
+    private ObjectStorageImpl objectStorage;
+
+    @Before
+    public void setup() throws IOException {
+
+        fileSystemTestingUtils.setup();
+        sessionInfo = mockSessionInfo();
+        fileSystem = mockFileSystem();
+        ioService = mockIoService( fileSystem );
+        objectStorage = new ObjectStorageImpl( ioService, "/object/store" );
```
ederign: please use the notation default://fs_name

uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/io/object/ObjectStorageImpl.java
ederign: @adrielparedes one way to do it is to make the ObjectStore a cdi bean, and make a manual lookup inside Jgit class.

porcelli: I don't think you will be able to use it on jgit underline

uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/io/object/ObjectStorageImpl.java
paulovmr: I want to give an example of what I said before: if I have three classes:

```
class MyClass {
   MyOtherClass myOtherClass;
}

class MyOtherClass { [...] }

class MyOtherClass2 extends MyOtherClass { [...] }
```

It's a requirement for the preferences stuff that:

```
MyClass myClass = new MyClass();
myClass.myOtherClass = new MyOtherClass2();
// serialization code here
myClass = // deserialization code here
if (myClass.myOtherClass instanceof MyOtherClass2) {
    // All good
} else {
    // Bad :(
}
```

@mrietveld @porcelli @ederign If any of you have experience with Gson and Kryo, do you know if that works? I can make some tests otherwise.

adrielparedes: @paulovmr let me test that! :)

paulovmr: @adrielparedes sure :)

mrietveld: @paulovmr AFAIK, Inheritance should be a non-issue for any serialization framework -- Kryo, and any XML or JSON serialization framework will all "capture" all type and field information. Whether or not `myClass.myOtherClass` is an instance of `MyOtherClass2` is just a test of the java code, not of serialization.  

Have you already seen problems related to this?

ederign: Just do place here the #uberfire conversation.
In order to do that you will have to write something similar to this:
https://gist.github.com/anonymous/e0f0b96dd83515acf23ef2bccfb189af#file-sample-java-L21
from here:
https://github.com/google/gson/blob/master/extras/src/main/java/com/google/gson/typeadapters/RuntimeTypeAdapterFactory.java

adrielparedes: I've tried with the GSON + TypeAdapter Strategy and things started to get weird. I'm completely agree that JSON is a much better alternative for communication stuffs, but if we need to store things, for that case I think that XStream is our guy. 

It's true that XStream is a more verbose thatn JSON, but it can store the whole tree easily with the correct classes.

adrielparedes:
https://github.com/uberfire/uberfire/pull/466#discussion_r75323087

uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/io/object/ObjectStorageImpl.java
```diff
@@ -0,0 +1,113 @@
...
+
+import com.thoughtworks.xstream.XStream;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.uberfire.io.IOService;
+import org.uberfire.java.nio.file.FileSystem;
+import org.uberfire.java.nio.file.FileSystemAlreadyExistsException;
+import org.uberfire.java.nio.file.Path;
+
+@Dependent
```
porcelli: you don't need this, @Dependent is default to CDI

uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/io/object/ObjectStorageImpl.java
```diff
@@ -0,0 +1,113 @@
...
+    private static final Logger logger = LoggerFactory.getLogger( ObjectStorageImpl.class );
+
+    private IOService ioService;
+
+    private FileSystem fileSystem;
+
+    private final XStream xs = new XStream();
+
+    @Inject
+    public ObjectStorageImpl( @Named("ioStrategy") final IOService ioService ) {
```
porcelli: for configs you shouldn't use "ioStrategy", use rather "configIO".

uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/io/object/ObjectStorageImpl.java
ederign: @adrielparedes, just talked with @csadilek and we already have this feature in Errai! \o/

//Server side
ServerMarshalling.toJSON(  );
ServerMarshalling.fromJSON(  );

//Client side
Marshalling.toJSON(  );
Marshalling.fromJSON(  );

@paulovmr already tested it and it fits his needs.

Could you please make this change?

Thanks

adrielparedes:
@paulovmr @ederign the code is up! please check if is it everything ok

paulovmr:
Looks good to me! +1 to merge IMO (when the build is green).

uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/io/object/ObjectStorageImpl.java
```diff
@@ -0,0 +1,113 @@
...
+
+public class ObjectStorageImpl implements ObjectStorage {
+
+    private static final Logger logger = LoggerFactory.getLogger( ObjectStorageImpl.class );
+
+    private IOService ioService;
+
+    private FileSystem fileSystem;
+
+    private final XStream xs = new XStream();
```
csadilek: This is needed anymore, right (and the import)?

uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/io/object/ObjectStorage.java
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * Read the content of the path given
+     * @param path The path where the content is stored
+     * @return The content of the file unmarshalled
+     */
+    <T> T read( String path );
+
+    /**
+     * Write an object into the object storage. That object is marshelled into XStream.
```
csadilek: No longer XStream....

uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/io/object/ObjectStorage.java
```diff
@@ -0,0 +1,70 @@
...
+     * Initialize the object storage.
+     * @param rootPath The root path where is going to create the file system
+     */
+    void init( String rootPath );
+
+    /**
+     * Checks if a path exists into the object storage
+     * @param path
+     * @return
+     */
```
csadilek: would be nice to describe ```@param``` and ```@return```...

adrielparedes:
@paulovmr @ederign @csadilek up again!

csadilek:
+1

uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/io/object/ObjectStorageImpl.java
```diff
@@ -0,0 +1,110 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.uberfire.io.IOService;
+import org.uberfire.java.nio.file.FileSystem;
+import org.uberfire.java.nio.file.FileSystemAlreadyExistsException;
+import org.uberfire.java.nio.file.Path;
+
+public class ObjectStorageImpl implements ObjectStorage {
+
+    private static final Logger logger = LoggerFactory.getLogger( ObjectStorageImpl.class );
```
ederign: Not used.

ederign:
Thanks @adrielparedes +1.

PR #467, opened 2016-08-17
NamedPosition should be handled as a Position (instead of a CompassPosition)

karreiro:
[This approach](https://github.com/uberfire/uberfire/blob/b3cc581826e5726cbfd350ab40fa53618b554d57/uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/workbench/PanelManagerImpl.java#L447) for converting the `position` will fail when it's a `NamedPosition`.

This PR solves the `kie-wb-distributions` tests. :-)

### DIFF
--- a/uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/workbench/PanelManagerImpl.java
+++ b/uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/workbench/PanelManagerImpl.java
@@ -442,12 +442,11 @@ public class PanelManagerImpl implements PanelManager {
         }
 
         PanelDefinition newPanel;
-        // Position instance could come from a different script so we convert
-        // it to the local type here first in order for == to work.
-        CompassPosition cp = CompassPosition.valueOf( "" + position );
-        if ( cp == CompassPosition.ROOT ) {
+
+        // Position instance could come from a different script so we compare using position.getName
+        if ( CompassPosition.ROOT.getName().equals( position.getName() ) ) {
             newPanel = rootPanelDef;
-        } else if ( cp == CompassPosition.SELF ) {
+        } else if ( CompassPosition.SELF.getName().equals( position.getName() ) ) {
             newPanel = targetPanelPresenter.getDefinition();
         } else {
             String defaultChildType = targetPanelPresenter.getDefaultChildType();
@@ -465,7 +464,7 @@ public class PanelManagerImpl implements PanelManager {
                                                childPanelPresenter );
 
             targetPanelPresenter.addPanel( childPanelPresenter,
-                                           cp );
+                                           position );
             newPanel = childPanel;
         }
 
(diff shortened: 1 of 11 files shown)


karreiro:
@csadilek Could you please review this PR? :-)

ederign:
@karreiro could you please give me a little more context about this?

csadilek:
@karreiro @ederign the code that's removed in this PR is needed for dynamic plugins to work (see source comment). Let's chat about this?

karreiro:
Yes @ederign :-) Some Selenium tests are broken on `kie-wb-distributions` with this [error](http://209.132.179.119:8080/job/kie-wb-distributions-pullrequests/319/testReport/org.kie.wb.selenium.ui/LoadAllPerspectivesIntegrationTest/allPerspectivesCanBeLoaded/): `Caused by: org.openqa.selenium.NoSuchElementException: Unable to locate element: {"method":"xpath","selector":"//div[contains(text(), 'Commits per organizational unit')]"}`

The perspective is not being loading correctly by the `PanelManager` when an error occurs in `position` convertion.

csadilek:
@karreiro @ederign if this method also needs to support named positions, unfortunately, the only way to fix it is to have a getName method in Position that returns a String and to compare the names here instead. This is needed so that the comparison will work across different scripts.

csadilek:
@karreiro @ederign 

I prepared a gist here with changes to Position/CompassPosition and NamedPosition: https://gist.github.com/csadilek/6f68c164c8f67432e5039decf2665b4b

Also, note the deprecation of getFieldName in favour of getName (see JavaDoc).

Then in PanelManager we can simply compare the names:
https://gist.github.com/csadilek/85ad7b8a7a361462542d4d1897d2d3c2

This should fix the issue and still work across different scripts.

karreiro:
Thanks @csadilek :-)

csadilek:
@karreiro thank you. Removing ```getFieldName``` will require rebuilding/deploying all workbench modules though as it's used in generated code which is packaged with jars. Not a big problem but something to keep in mind. Probably best if I deployed snapshots of drools-wb, jbpm-console once we merged this.

uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/workbench/PanelManagerImpl.java
```diff
@@ -442,12 +442,11 @@ public PanelDefinition addWorkbenchPanel( final PanelDefinition targetPanel,
         }
 
         PanelDefinition newPanel;
-        // Position instance could come from a different script so we convert
-        // it to the local type here first in order for == to work.
-        CompassPosition cp = CompassPosition.valueOf( "" + position );
-        if ( cp == CompassPosition.ROOT ) {
+
+        // Position instance could come from a different script so we compare using position.getName
+        if ( position.getName().equals( CompassPosition.ROOT.getName() ) ) {
```
csadilek: In general CompassPosition.ROOT.getName().equals(position.getName()) is better as it's safe even if position or position.getName is null.

csadilek:
+1

csadilek:
I will merge this and take care of deploying snapshots.

PR #473, opened 2016-08-22
Byteman test modified due an unexpected behaviour

adrielparedes:
@ederign the test was commented because there is an unexpected behavior that locks the other test into that class. This is a workaround.


uberfire-nio2-backport/uberfire-nio2-impls/uberfire-nio2-jgit/src/test/java/org/uberfire/java/nio/fs/jgit/JGitFileSystemProviderBytemanTest.java
```diff
@@ -50,7 +50,7 @@
 
     private static Logger logger = LoggerFactory.getLogger( JGitFileSystemProviderBytemanTest.class );
 
-    @Test()
+    //    @Test()
```
psiroky: Nit-picking, but you could use @Ignore annotation (which also accepts string explaining why is the test ignored). The advantage of this approach is that the test gets reported as skipped by jUnit (which is visible in the test report).

adrielparedes: @psiroky let me chage it, It's going to take me a minute

ederign:
+1 for this while @adrielparedes find the real issue.

PR #474, opened 2016-08-23
[RHBPMS-4105] concurrent JGit test

tkobayas:
Concurrent JGitUtil access test.
Thread 1 : Keep committing new files to a repo
Thread 2 : Keep reading (resolvePath) an existing file in the repo.
It runs successfully in Linux (Fedora 22) but it constantly fails in Windows 7.

The loop counts (200 for committer thread. 2000 for reader thread) are arbitrary values that I can reproduce the issue constantly but try to minimize the test time. You may need to increase the values to reproduce the issue in a faster Windows machine.


ederign:
Thanks @tkobayas. I'll take a look on this.

porcelli:
-1 for it; it should use byteman to force concurrency issue

PR #475, opened 2016-08-23
Byteman test locking solved

adrielparedes:
@ederign It was not running in parallel, but it was getting stuck because of the repeatead rendezvous name. I've changed it and now is working properly


ederign:
thanks @adrielparedes! +1

PR #476, opened 2016-08-29
UF-396: Inherit qualifiers in generated activities

paulovmr:
@csadilek @ederign Hi guys, do you mind reviewing it? This PR also removes the previously created WorkbenchPreferencesActivity, because it won't be needed anymore.

Obs.: Apparently Git resolved to show totally different files as "moved" instead of separately "removed" and "added", this might make the review a bit more confusing...


uberfire-workbench/uberfire-workbench-processors/src/main/java/org/uberfire/annotations/processors/GeneratorUtils.java
```diff
@@ -1349,4 +1350,54 @@ static AnnotationValue extractAnnotationPropertyValue( Elements elementUtils,
...
     }
+
+    /**
+     * This method builds an annotation source-code declaration based on an {@link AnnotationMirror}.
+     * The supported annotation field types are Class, String and any type which its toString() method
+     * corresponds to its value declaration (e.g. boolean, int).
+     * @param annotation {@link AnnotationMirror} which the declaration will be based on.
+     * @return The annotation source-code declaration.
+     */
+    public static String buildAnnotationDeclaration( final AnnotationMirror annotation ) {
```
csadilek: @paulovmr couldn't you just use annotationMirror.toString()? Just played around with it a bit and it seems to return a valid declaration string.

csadilek:
+1 (one minor question inline)

uberfire-workbench/uberfire-workbench-processors/src/main/java/org/uberfire/annotations/processors/GeneratorUtils.java
paulovmr: @csadilek Just tried it here and it works! Much better this way, I'm changing it here right now. Thanks!

ederign:
Thanks @paulovmr ! :)

PR #478, opened 2016-08-31
GUVNOR-2654: Messages panel not visible

csadilek:
@ederign @manstis  can you review/merge this please to fix the QE tests, see: https://issues.jboss.org/browse/GUVNOR-2654

Basically a panel definition could come from an external script, so I should've added this to begin with (see JavaDocs of PluginUtil). In this case (same script) it was just a mismatch of dealing with the default values (null, -1).


ederign:
+1 thanks @csadilek

PR #482, opened 2016-09-02
latest Zanata translation changes in uberfire from 2016-09-02

kiereleaseuser:


mbiarnes:
same content as PR #485 - so this PR can be closed

PR #437, opened 2016-07-11
Uf docs errai4

ederign:
@paulovmr @karreiro as you guys  were the last one's that made the tutorial, could you please review it?

I've removed the gwt widgets from all screens.

(perspectives should be addressed after this ticket https://issues.jboss.org/browse/UF-358)

PR #483, opened 2016-09-02
RHBRMS-349: New Item Menu: No Project selected, all items enabled after Perspective reset

manstis:
See https://issues.jboss.org/browse/RHBRMS-349

The cause appears to be when a ```PerspectiveChange``` event is fired the ```WorkbenchMenuBarPresenter``` calls the Perspective's ```getMenus()``` method. These methods *could* set the ```MenuItems``` enabled/disabled state however when the menu is passed to the ```View``` the UI Elements' state is not synchronized with the ```MenuItem```. This fix forces the ```MenuItem``` state to be reflected in the UI Elements.

### DIFF
--- a/uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/workbench/widgets/menu/WorkbenchMenuBarPresenter.java
+++ b/uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/workbench/widgets/menu/WorkbenchMenuBarPresenter.java
@@ -19,12 +19,13 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
-
 import javax.annotation.PostConstruct;
 import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.event.Observes;
 import javax.inject.Inject;
 
+import com.google.gwt.user.client.ui.IsWidget;
+import com.google.gwt.user.client.ui.Widget;
 import org.jboss.errai.ioc.client.container.IOC;
 import org.jboss.errai.security.shared.api.identity.User;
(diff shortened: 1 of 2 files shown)


uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/workbench/widgets/menu/WorkbenchMenuBarPresenter.java
```diff
@@ -261,9 +259,9 @@ public void enabledStateChanged( final boolean enabled ) {
                 } );
             }
 
-        }
+        } ) );
 
-        ) );
+        synchronizeUIWithMenus( addedMenu.getItems() );
```
manstis: This is the "guts" of the change.. everything else is formatting....

uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/workbench/widgets/menu/WorkbenchMenuBarPresenter.java
```diff
@@ -274,70 +272,72 @@ private void addPerspectiveMenus( final PerspectiveActivity perspective ) {
...
+                                                        enabled );
                         }
-                    });
+                    } );
                 }
 
-            }));
+            } ) );
+
+            synchronizeUIWithMenus( menus.getItems() );
```
manstis: This is the "guts" of the change.. everything else is formatting....

manstis:
Part of an ensemble:

- uberfire/uberfire#483
- #437

manstis:
@ederign @cristianonicolai @paulovmr Mind taking a look?

ederign:
thanks @manstis +1

PR #486, opened 2016-09-05
RHBRMS-120: Business Central's RepositoryEditor view does not reflect the correct Git/SSH port number if they have been changed while starting the server

manstis:
See https://issues.jboss.org/browse/RHBRMS-120

The System Properties removed are:-

1. Not mentioned in the [User Guide](http://docs.jboss.org/drools/release/6.4.0.Final/drools-docs/html/ch18.html#wb.systemProperties)
2. Not used by either the SSH or Git daemons.
3. Only used by the UI related code to construct the URLs shown to Users.

Using ```daemonHostAddr```, ```daemonPort```, ```sshHostAddr``` and ```sshPort``` makes what is shown to Users consistent with the actual daemon configuration.


manstis:
@porcelli Fancy looking at some old code?

porcelli:
Concerned that we might not have a DNS friendly output, always referencing IP numbers instead of the machine name.

manstis:
@porcelli Yes.. however https://issues.jboss.org/browse/RHBPMS-4126 asks for IP address and not Host Name!! (Furthermore - see my comment on RHBPMS-4126 - we probably reliably determine the Host Name or IP address "visible" to the User anyway.. perhaps I'll add back the System Property for Host Name so that can be set if required. The fix for port number is still required though.

porcelli:
"add back the System Property for Host Name so that can be set if required. The fix for port number is still required though." -> +1

ederign:
+1 thanks @manstis

PR #494, opened 2016-09-06
Avoid resizing dock panels if they're not setup yet ( do NOT merge yet, only a possible fix suggested )

romartin:
Hi @ederign , @paulovmr !

As commented today on my IRC, just got back from PTO I tried to run the new designer's showcase (https://github.com/romartin/wirez) and I get a client error that was not present two weeks ago, I had no changes on the project, so not sure if externally something has changed on last weeks or I'm missing something from my side.

The problem is that just after loading the home it appears an error popup saying that "heights  cannot be negative", and trying to get deeper I get a negative value for "height", as you can see  here http://picpaste.com/pics/Screenshot_from_2016-09-06_23-01-53-RZ1IsYlB.1473202154.png

Not sure why yet but the method DocksExpandedBar#onResize  is called by a gwt client scheduler BEFORE the call for DocksExpandedBar#setup, so when trying to resize title is not yet set... 

This PR provides a really quick fix that solves at least my issue, not sure if makes sense at 100%, so what do you think? After this fix all works nice.

Thanks!

### DIFF
--- a/uberfire-extensions/uberfire-simple-docks/uberfire-simple-docks-client/src/main/java/org/uberfire/client/docks/view/bars/DocksExpandedBar.java
+++ b/uberfire-extensions/uberfire-simple-docks/uberfire-simple-docks-client/src/main/java/org/uberfire/client/docks/view/bars/DocksExpandedBar.java
@@ -20,15 +20,9 @@ import com.google.gwt.core.client.GWT;
 import com.google.gwt.core.client.Scheduler;
 import com.google.gwt.event.dom.client.ClickEvent;
 import com.google.gwt.event.dom.client.ClickHandler;
-import com.google.gwt.event.logical.shared.ResizeEvent;
-import com.google.gwt.event.logical.shared.ResizeHandler;
 import com.google.gwt.uibinder.client.UiBinder;
 import com.google.gwt.uibinder.client.UiField;
-import com.google.gwt.user.client.ui.Composite;
-import com.google.gwt.user.client.ui.FlowPanel;
-import com.google.gwt.user.client.ui.ProvidesResize;
-import com.google.gwt.user.client.ui.RequiresResize;
-import com.google.gwt.user.client.ui.Widget;
+import com.google.gwt.user.client.ui.*;
 import org.gwtbootstrap3.client.ui.Button;
 import org.gwtbootstrap3.client.ui.ButtonGroup;
 import org.gwtbootstrap3.client.ui.Heading;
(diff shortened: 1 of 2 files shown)


paulovmr:
Hi @romartin , I think it's a fair change, since you cannot resize the dock unless the title is present. Just one thing, could you add a test to cover this logic?

romartin:
Hi @paulovmr , @ederign !

I have opened the issue https://issues.jboss.org/browse/UF-400 for it. Added a new commit with the test case for it as well.

Thanks!

ederign:
@romartin thanks for your PR. See my comments inline.

uberfire-extensions/uberfire-simple-docks/uberfire-simple-docks-client/src/main/java/org/uberfire/client/docks/view/bars/DocksExpandedBar.java
```diff
@@ -182,4 +164,23 @@ public void clear() {
...
     }
+
+    public void setPanelSize( final int width,
+                              final int height ) {
+        targetPanel.setPixelSize( width, height );
+    }
+
+    void resizeTargetPanel() {
+
+        final int height = getOffsetHeight() - ( isTitleSet() ? titlePanel.getOffsetHeight() : 0 );
```
ederign: After initialization, Uberfire Docks fire this event: 
https://github.com/ederign/uberfire/blob/f8e238c0a9c2599904e64f84e5acd3da06c4701b/uberfire-extensions/uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/breadcrumbs/UberfireBreadcrumbs.java#L79

What do you think of only allow resize after this?

romartin: Adding some kind of flag you mean? 
Anyway it's quite strange this because as far as I understand the method setup here  should be called before the resize one, so title shouldn't be null... at least empty, isn't it? because probably does not makes sense taking into account the titlePanel size if no title is set. So allowing resize after the event is fired seems fine! But this logic about the title panel makes sense as well? what you think? thx @ederign

ederign:
@romartin the issue is fixed [here](https://github.com/uberfire/uberfire/pull/506)! Thanks for the debug and testing! :)

romartin:
@ederign Thanks!!

PR #499, opened 2016-09-08
UF-401: improve for random windows fails.

porcelli:
Tests are now using byteman instead of lucky thread.


ederign:
@porcelli thanks for this?

Maybe change Supplier for another name to avoid clash with java util Supplier.

I'll try this on Windows right now.

porcelli:
it doesn't clash, packages are exactly for that :)

but it's fine.

ederign:
+1. @porcelli works also on windows (with old GSS test).

PR #501, opened 2016-09-09
RHBPMS-4190 : [GSS](6.3.z) Custom changes to pom.xml are being overwritten

Rikkola:
https://issues.jboss.org/browse/RHBPMS-4190

Not sure what the identity check originally protected against. My first subject was that the reload() might reload the active editor, but this does not happen.

### DIFF
--- a/uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/mvp/LockManagerImpl.java
+++ b/uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/mvp/LockManagerImpl.java
@@ -296,9 +296,7 @@ public class LockManagerImpl implements LockManager {
 
     void onResourceUpdated( @Observes ResourceUpdatedEvent res ) {
         if ( lockTarget != null && res.getPath().equals( lockTarget.getPath() ) ) {
-            if ( !res.getSessionInfo().getIdentity().equals( user ) ) {
-                reload();
-            }
+            reload();
             releaseLock();
         }
     }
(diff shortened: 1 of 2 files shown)


manstis:
@Rikkola this is probably one for @csadilek to review.

```ResourceUpdatedEvent``` is fired from VFS when a file is updated. The (deleted) check for ```user``` would ensure the editor was only reloaded for *other* Users and not the current User; thus AFAIK giving the *other* Users the opportunity to acquire the lock..

Rikkola:
Could be, have to say I am not 100% sure.
What I think this does is just refresh the editor if any changes to the file happened, while any user has the file open. Locked or not.

Previously it only refreshed the editor for other users.

csadilek:
@Rikkola @manstis The problem with removing this check is that now when a user saves a file their editor will reload although it already has the latest version. That's definitely not desirable from a UX perspective. I think we should solve this differently in the project "editor" or come up with an enhanced check to see if the file was edited by the same user but by another editor.

csadilek:
@Rikkola are you sure the reload doesn't happen for the active editor after you save? I definitely remember it doing that. Don't see why it wouldn't still.

csadilek:
@Rikkola @manstis I've confirmed that removing this check causes problems :(. The reload does then also happen for the active user/editor which causes a couple of issues e.g. when you edit and save metadata the reload forces you back to the main tab (instead of staying on the overview tab which you just used).

The "problem" described in the JIRA is minor and unrelated to asset locking. It was always there and is caused by the fact that there are two ways of editing a pom.xml file, a real editor and the project editor/screen. I can think of a couple of way how to solve that, but I think we should discuss this.

For this PR, -1, sorry, as the fix causes a bigger issue than the one it fixes in project editor.

Rikkola:
@csadilek You make me doubt that it does not. Just tested with Project Editor, so could be that the base editor does something weird.
Also, like I said before. The fact that the check for this is in, makes me think this breaks something.

PR #502, opened 2016-09-09
[0.9.x] RHBPMS-4190 : [GSS](6.3.z) Custom changes to pom.xml are being overwritten

Rikkola:
Same as for master.

(cherry picked from commit b9bbdcd86de3c9814a2a49b5d43ab450a8bb059c)


csadilek:
@Rikkola and I had a chat and came up with a better solution. This problem happens because a single user can use multiple editors to change the same asset where one of the editors is not a KIEeditor and doesn't have conflict resolution strategies (the project editor is a screen).

This is not related to UF locking as it only is an issue if a single user works on one asset it multiple editors. /cc @manstis @ederign

PR #503, opened 2016-09-09
UF-380: Perspective plugin html component doesn't display new content

paulovmr:


ederign:
Hi @paulovmr could you please add some tests?

paulovmr:
I've added a test. Thanks, @ederign !

ederign:
+1 thanks.

PR #505, opened 2016-09-13
GUVNOR-2680: VersionRecordManager: Index Out of Bounds when opening a deleted file

manstis:
See https://issues.jboss.org/browse/GUVNOR-2680

@Rikkola could you please review. I may have fixed a symptom not the cause (as presumably ```VersionService``` may have used to always return at least one ```VersionRecord```). It could be the regression was introduced with the refactoring done for asset vs meta-data ```VersionRecords```.

If this "patch" is of a symptom and not the cause feel free to say!


manstis:
Jenkins please retest this.

Rikkola:
I guess it makes sense that the file has not `VersionRecord` if it is deleted.
With Git it would be possible to get the full history including the commit that deleted the file, but I guess we should not expose too much of the Git underneath. Git is pretty much the only file system type that could support this.

Rikkola:
Again nice clean fix.

+1

ederign:
Thanks @manstis

PR #508, opened 2016-09-15
UF-407: Added support for proxy authentication using http[s].proxyUser

porcelli:
and http[s].proxyPassword


ederign:
+1 thanks @porcelli

PR #515, opened 2016-09-21
UF-386: Workbench Preferences Bean API

paulovmr:


uberfire-extensions/uberfire-preferences/uberfire-preferences-api/src/main/java/org/uberfire/ext/preferences/shared/PropertyFormType.java
```diff
@@ -0,0 +1,79 @@
...
+
+    TEXT {
+        @Override
+        public Object fromString( final String stringValue ) {
+            return stringValue;
+        }
+
+        @Override
+        public String toString( final Object realValue ) {
+            return (String) realValue;
```
manstis: Would it be better to call ```realValue.toString()```?

manstis: Or ```String.valueOf( o )```?

uberfire-extensions/uberfire-preferences/uberfire-preferences-api/src/main/java/org/uberfire/ext/preferences/shared/PropertyFormType.java
```diff
@@ -0,0 +1,79 @@
...
+        }
+    }, SECRET_TEXT {
+        @Override
+        public Object fromString( final String stringValue ) {
+            return stringValue;
+        }
+
+        @Override
+        public String toString( final Object realValue ) {
+            return (String) realValue;
```
manstis: ```String.valueOf( o )```?

uberfire-extensions/uberfire-preferences/uberfire-preferences-api/src/main/java/org/uberfire/ext/preferences/shared/PropertyFormType.java
```diff
@@ -0,0 +1,79 @@
...
+        }
+    }, COLOR {
+        @Override
+        public Object fromString( final String stringValue ) {
+            return stringValue;
+        }
+
+        @Override
+        public String toString( final Object realValue ) {
+            return (String) realValue;
```
manstis: ```String.valueOf( o )```?

uberfire-extensions/uberfire-preferences/uberfire-preferences-api/src/main/java/org/uberfire/ext/preferences/shared/annotations/WorkbenchPreference.java
```diff
@@ -0,0 +1,47 @@
...
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.TYPE })
+public @interface WorkbenchPreference {
+
+    /**
+     * Defines if a preference is a root preference, which means it will appear in the first
+     * level of the hierarchic structure printed in the configuration screen.
+     * @return True if it is a root preference, and false otherwise.
+     */
+    boolean root() default false;
```
manstis: If the default is ```false``` could a preference omit to be listed anywhere?! Would a default of ```true``` be better - at least then a developer would see the preference listed at root level and think "feck.. it's meant to be a sibling!"

uberfire-extensions/uberfire-preferences/uberfire-preferences-api/src/main/java/org/uberfire/ext/preferences/shared/bean/BasePreference.java
```diff
@@ -0,0 +1,73 @@
...
+import org.uberfire.mvp.Command;
+import org.uberfire.mvp.ParameterizedCommand;
+
+/**
+ * Interface that all preference beans should implement. It allows load and save operations
+ * when injecting it through CDI. The methods in this class are only really implemented by
+ * a generated preference bean, that will be the one injected by CDI.
+ * @param <T> The preference bean type implementing the interface.
+ */
+public interface BasePreference<T> extends Preference {
```
manstis: Do we really need ```default``` implementations of a "green field" interface (i.e. it's unlikely existing code implements this and you don't want to break compatibility).

uberfire-extensions/uberfire-preferences/uberfire-preferences-api/src/main/java/org/uberfire/ext/preferences/shared/bean/PreferenceBeanServerStore.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.Collection;
+import java.util.List;
+
+import org.jboss.errai.bus.server.annotations.Remote;
+
+/**
+ * Service to manage preference beans, on the server-side.
+ */
+@Remote
+public interface PreferenceBeanServerStore extends PreferenceBeanStore {
```
manstis: ```PreferenceBeanStore``` is documented as "Service to manage preference beans, on the server-side **and client-side**" and is annotated ```@Remote```. This interface is also annotated ```@Remote``` but documented as "Service to manage preference beans, on the server-side" (i.e. no mention of "client"). Since it is annotated ```@Remote``` it'll also be available client-side. Is the documentation wrong or the annotation?

uberfire-extensions/uberfire-preferences/uberfire-preferences-api/src/main/java/org/uberfire/ext/preferences/shared/bean/PreferenceHierarchyElement.java
```diff
@@ -0,0 +1,136 @@
...
+
+import org.jboss.errai.common.client.api.annotations.MapsTo;
+import org.jboss.errai.common.client.api.annotations.Portable;
+
+/**
+ * Holds all preference information necessary to build a hierarchic interface for preferences.
+ * @param <T> Preference bean type.
+ */
+@Portable
+public class PreferenceHierarchyElement<T> {
```
manstis: Is there any value in adding ```hashCode()``` and ```equals()```? "Identical" instances marshalled from the server to the client would not pass the default (identity) equality test of ```Object.equals()``` - this comment applies to all of your ```@Portable``` classes.

uberfire-extensions/uberfire-preferences/uberfire-preferences-backend/src/main/java/org/uberfire/ext/preferences/backend/PreferenceBeanStoreImpl.java
```diff
@@ -0,0 +1,322 @@
...
+import org.uberfire.ext.preferences.shared.bean.PreferenceBeanStore;
+import org.uberfire.ext.preferences.shared.bean.PreferenceHierarchyElement;
+import org.uberfire.mvp.Command;
+import org.uberfire.mvp.ParameterizedCommand;
+
+/**
+ * Backend implementation for {@link PreferenceBeanStore}.
+ */
+@Service
+public class PreferenceBeanStoreImpl implements PreferenceBeanServerStore {
```
manstis: Out of interest, this class appears to have both what is considered "client" API and "server" API. Since you have a ```@Remote``` interface for the client; and a "normal" server-side implementation annotated ```@Service``` why do we need to explicitly provide "success" and "failure" callbacks rather than use Errai's ```Caller``` mechanism to handle success and failure states? I'm just interested to understand really.

uberfire-extensions/uberfire-preferences/uberfire-preferences-client/src/main/java/org/uberfire/ext/preferences/client/annotations/PreferenceForm.java
```diff
@@ -0,0 +1,42 @@
...
+@Inherited
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.TYPE, ElementType.FIELD, ElementType.METHOD, ElementType.PARAMETER })
+public @interface PreferenceForm {
+
+    /**
+     * Preference bean class that the form edits.
+     * @return Preference bean class.
+     */
+    Class<?> value() default void.class;
```
manstis: Should this not be ```Classs<? extends BasePreferenceForm>```?

uberfire-extensions/uberfire-preferences/uberfire-preferences-client/src/main/java/org/uberfire/ext/preferences/client/event/HierarchyItemSelectedEvent.java
```diff
@@ -0,0 +1,52 @@
...
+
+import java.util.Map;
+
+import org.uberfire.ext.preferences.shared.bean.BasePreferencePortable;
+import org.uberfire.ext.preferences.shared.bean.PreferenceHierarchyElement;
+
+/**
+ * Event fired when an preference hierarchy item is selected at the preferences central perspective.
+ */
+public class HierarchyItemSelectedEvent extends AbstractHierarchyItemEvent {
```
manstis: What's the difference between ```HierarchyItemFormInitializationEvent``` and this class? Both appear to provide information about a Preference when it is selected in the "Preferences Central Perspective" - also, as an aside, I assume "Preferences" UI is a ```@WorkbenchScreen``` that could be used from any perspective (in case UX decide, for example, to want to show under "Admin" or somewhere).

uberfire-extensions/uberfire-preferences/uberfire-preferences-client/src/main/java/org/uberfire/ext/preferences/client/utils/PreferenceFormBeansInfo.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    /**
+     * Searches for a {@link WorkbenchScreenActivity} that is qualified with {@link PreferenceForm} with the passed
+     * preference bean class as its value.
+     * @param preferenceBeanClass The {@link PreferenceForm} value to be searched.
+     * @return The screen bean identifier.
+     */
+    public String getPreferenceFormFor( Class<?> preferenceBeanClass ) {
+        final Collection<? extends IOCBeanDef<?>> screenBeans = activityBeansInfo.lookupBeans( WorkbenchScreenActivity.class );
```
manstis: Do ```screenBeans``` need to extend ```BasePreferenceForm```? You might be able to narrow down the generics.

uberfire-extensions/uberfire-preferences/uberfire-preferences-processors/src/main/resources/org/uberfire/ext/preferences/processors/templates/workbenchPreferenceBean.ftl
```diff
@@ -0,0 +1,94 @@
...
+import javax.inject.Inject;
+
+import org.uberfire.ext.preferences.shared.bean.BasePreferenceBean;
+import org.uberfire.ext.preferences.shared.bean.BasePreferencePortable;
+import org.uberfire.ext.preferences.shared.bean.PreferenceBeanStore;
+import org.uberfire.mvp.Command;
+import org.uberfire.mvp.ParameterizedCommand;
+
+@Dependent
+public class ${targetClassName} extends ${sourceClassName} implements BasePreferenceBean<${sourceClassName}> {
```
manstis: We have customarily put the ```@Generated``` [annotation](https://github.com/uberfire/uberfire/blob/master/uberfire-workbench/uberfire-workbench-processors/src/main/resources/org/uberfire/annotations/processors/templates/activityEditor.ftl#L76) these templates, but it's your choice!

uberfire-extensions/uberfire-wires/uberfire-wires-webapp/src/main/java/org/uberfire/ext/wires/client/preferences/central/form/DefaultPreferenceFormView.java
```diff
@@ -0,0 +1,56 @@
...
+                                                  IsElement {
+
+    private DefaultPreferenceForm presenter;
+
+    @Inject
+    @DataField("properties-editor")
+    PropertyEditorWidget propertiesEditorWidget;
+
+    /*@DataField("properties-editor")
+    Element propertiesEditor = DOM.createDiv();*/
```
manstis: I assume you was trying to make this ```Element``` based (instead of ```Widget```)? If the commented code blocks are not needed perhaps it'd be an idea to remove them?

uberfire-extensions/uberfire-preferences/uberfire-preferences-api/src/main/java/org/uberfire/ext/preferences/shared/PropertyFormType.java
paulovmr: I agree, will change that.

uberfire-extensions/uberfire-preferences/uberfire-preferences-api/src/main/java/org/uberfire/ext/preferences/shared/PropertyFormType.java
paulovmr: I agree, will change that.

uberfire-extensions/uberfire-preferences/uberfire-preferences-api/src/main/java/org/uberfire/ext/preferences/shared/PropertyFormType.java
paulovmr: I agree, will change that.

uberfire-extensions/uberfire-preferences/uberfire-preferences-api/src/main/java/org/uberfire/ext/preferences/shared/annotations/WorkbenchPreference.java
paulovmr: I agree, will change that.

uberfire-extensions/uberfire-preferences/uberfire-preferences-api/src/main/java/org/uberfire/ext/preferences/shared/bean/BasePreference.java
paulovmr: I haven't liked that very much either, but I couldn't think in a better way to do it. The reason to provide these default implementations is so that the user can inject its preference bean anywhere, and call those methods (save and load) without having to provide an implementation for them. The real implementation will be made by generated code (SomePrefererenceBeanGeneratedImpl extends SomePreference), which will be the type of the instance provided by CDI when a preference bean is injected.

uberfire-extensions/uberfire-preferences/uberfire-preferences-backend/src/main/java/org/uberfire/ext/preferences/backend/PreferenceBeanStoreImpl.java
paulovmr: This was done this way because this service will be called from a generated shared code, since preference beans can be injected on client and server code. So it has to work on both client and server. Since the generated code is the same for both, I had to leave to CDI to inject whichever instance is appropriated, but with the same interface. That's why the server service ended up with some callback methods. I figured it shouldn't be a problem, since only generated code should be calling this service anyway.

uberfire-extensions/uberfire-preferences/uberfire-preferences-client/src/main/java/org/uberfire/ext/preferences/client/utils/PreferenceFormBeansInfo.java
paulovmr: The user's `@WorkbenchScreen` classes will have to extend `BasePreferenceForm`, but the generated activity (the one I'm looking for) won't extend it. Not sure if that's what you meant.

uberfire-extensions/uberfire-preferences/uberfire-preferences-processors/src/main/resources/org/uberfire/ext/preferences/processors/templates/workbenchPreferenceBean.ftl
paulovmr: I agree, will change that.

uberfire-extensions/uberfire-wires/uberfire-wires-webapp/src/main/java/org/uberfire/ext/wires/client/preferences/central/form/DefaultPreferenceFormView.java
paulovmr: Yes, forgot that. :/

uberfire-extensions/uberfire-preferences/uberfire-preferences-api/src/main/java/org/uberfire/ext/preferences/shared/bean/PreferenceBeanServerStore.java
paulovmr: The documentation and the `@Remote` annotation in `PreferenceBeanStore` are wrong, I will change them. Thanks.

uberfire-extensions/uberfire-preferences/uberfire-preferences-client/src/main/java/org/uberfire/ext/preferences/client/event/HierarchyItemSelectedEvent.java
paulovmr: They provide the same information, but they are fired and observed in different moments. The `HierarchyItemSelectedEvent` is fired when an item is selected in the hierarchy structure (so far, a tree) in the preferences central screen. Then it is observed and, according to the selected preference, a `@WorkbenchScreen`is called. Immediatly after that, a `HierarchyItemFormInitializationEvent` is fired, and observed by the default form or the `BasePreferenceForm` (if it is a custom form). Then the preference form gets access to the preference value. It had to be done this way (through events) because we can't pass complex arbitrary objects directly to a screen through the place manager.

uberfire-extensions/uberfire-preferences/uberfire-preferences-client/src/main/java/org/uberfire/ext/preferences/client/annotations/PreferenceForm.java
paulovmr: Actually it should be `Class<? extends BasePreference>`. I will change that.

paulovmr:
@manstis Thanks a lot for the review! I answered your comments and will start to apply the ones I agreed to. 

I also look forward to it, and will certainly need your help when we migrate the workbench to use this new feature :).

uberfire-extensions/uberfire-preferences/uberfire-preferences-api/src/main/java/org/uberfire/ext/preferences/shared/annotations/Property.java
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * Defines the type of field to be used in the default form. The default is TEXT.
+     * You should only specify this if you are using the default form for preferences
+     * edition, otherwise it will have no effect.
+     * @return The property type to be used.
+     */
+    PropertyFormType formType() default PropertyFormType.TEXT;
+
+    /**
```
csadilek: Defines whether or not this property should be inherited by sub types. Defaults to false.

uberfire-extensions/uberfire-preferences/uberfire-preferences-api/src/main/java/org/uberfire/ext/preferences/shared/annotations/WorkbenchPreference.java
csadilek: Wondering if we need this property at all. Shouldn't we be able to infer this at compile-time?

uberfire-extensions/uberfire-preferences/uberfire-preferences-api/src/main/java/org/uberfire/ext/preferences/shared/bean/BasePreferencePortable.java
```diff
@@ -0,0 +1,77 @@
...
+
+package org.uberfire.ext.preferences.shared.bean;
+
+import java.util.Map;
+
+import org.uberfire.ext.preferences.shared.PropertyFormType;
+import org.uberfire.ext.preferences.shared.annotations.WorkbenchPreference;
+
+/**
+ * Interface that all generated bean implementations of preference beans implement.
```
csadilek: Same Javadocs as BasePreferenceBean...

uberfire-extensions/uberfire-preferences/uberfire-preferences-api/src/main/java/org/uberfire/ext/preferences/shared/bean/Preference.java
```diff
@@ -0,0 +1,24 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.uberfire.ext.preferences.shared.bean;
+
+/**
+ * Interface that all preference beans and their generated implementations implement.
```
csadilek: Needs to describe what this interface represents, not that it is an interface and which classes implement it.

uberfire-extensions/uberfire-preferences/uberfire-preferences-backend/src/main/java/org/uberfire/ext/preferences/backend/PreferenceBeanStoreImpl.java
```diff
@@ -0,0 +1,322 @@
...
+
+    private static final AnnotationLiteral<org.uberfire.ext.preferences.shared.annotations.PortablePreference> portablePreferenceAnnotation = new AnnotationLiteral<org.uberfire.ext.preferences.shared.annotations.PortablePreference>() { };
```
csadilek: Why is this fully qualified?

uberfire-extensions/uberfire-preferences/uberfire-preferences-backend/src/main/java/org/uberfire/ext/preferences/backend/PreferenceBeanStoreImpl.java
```diff
@@ -0,0 +1,322 @@
...
+    }
+
+    private <U extends BasePreference<U>, T extends BasePreferencePortable<U>> void saveSubPreference( final Object portablePreference,
+                                                                                                       final Field field ) throws IllegalAccessException {
+        final Class<U> propertyType = (Class<U>) field.getType();
+        final T subPreferenceValue = (T) field.get( portablePreference );
+        save( propertyType, subPreferenceValue );
+    }
+
+    private <T extends BasePreference<T>> void saveOne( final BasePreferencePortable<?> portablePreference ) {
```
csadilek: Personally I wouldn't use saveOne and saveAll as names but use method overloading instead.

uberfire-extensions/uberfire-wires/uberfire-wires-webapp/src/main/java/org/uberfire/ext/wires/client/preferences/central/form/DefaultPreferenceForm.java
```diff
@@ -0,0 +1,165 @@
...
+        category.undo();
+        view.init( this );
+    }
+
+    public BasePreferencePortable<?> getPreference() {
+        return preference;
+    }
+
+    public PropertyEditorType getPropertyEditorType( PropertyFormType propertyFormType ) {
+        if ( PropertyFormType.TEXT.equals( propertyFormType ) ) {
```
csadilek: Hm is it worth having both enum types when there values always have to correspond? If it is, couldn't you use valueOf here?

csadilek:
@paulovmr One bigger thing we need to consider is that dynamic plugins should be able to provide ```WorkbenchPreferences```. That means a ```WorkbenchPreference``` defined in a dynamic plugin should be discovered by the main application. In order for that to happen the generated implementation needs to be a ```@JsType```...

uberfire-extensions/uberfire-preferences/uberfire-preferences-api/src/main/java/org/uberfire/ext/preferences/shared/bean/BasePreference.java
csadilek: Prob. better to throw a runtime exception if the unimplemented version is called...i.e. load is called on the portable generated impl.

uberfire-docs/docs/uberfireExtensions/preferencesBeansAPI.md
```diff
@@ -0,0 +1,120 @@
+# Beans API
+
+## Introduction
+
+The Preferences Beans API provides you a way to create POJOs and, by adding some configuration annotations, transform them in a preference object with out-of-the-box support to edition and management of their value by each user.
```
paulovmr: Done.

uberfire-extensions/uberfire-preferences/uberfire-preferences-api/src/main/java/org/uberfire/ext/preferences/shared/annotations/Property.java
paulovmr: Done.

uberfire-extensions/uberfire-preferences/uberfire-preferences-api/src/main/java/org/uberfire/ext/preferences/shared/bean/BasePreferencePortable.java
paulovmr: Fixed.

uberfire-extensions/uberfire-preferences/uberfire-preferences-api/src/main/java/org/uberfire/ext/preferences/shared/bean/Preference.java
paulovmr: Done.

uberfire-extensions/uberfire-preferences/uberfire-preferences-backend/src/main/java/org/uberfire/ext/preferences/backend/PreferenceBeanStoreImpl.java
paulovmr: Changed.

uberfire-extensions/uberfire-preferences/uberfire-preferences-backend/src/main/java/org/uberfire/ext/preferences/backend/PreferenceBeanStoreImpl.java
paulovmr: I've changed the `saveAll` to `save`, but the `saveOne` (private) conflicts with another `save` with one generic parameter, so I've kept the name.

uberfire-extensions/uberfire-wires/uberfire-wires-webapp/src/main/java/org/uberfire/ext/wires/client/preferences/central/form/DefaultPreferenceForm.java
paulovmr: I've made this way to not make the API depend on the properties editor, since we will probably change it to the form builder. But I agree with your suggestion, I've changed it to `valueOf`.

uberfire-extensions/uberfire-preferences/uberfire-preferences-api/src/main/java/org/uberfire/ext/preferences/shared/bean/BasePreference.java
paulovmr: I agree, done. :)

paulovmr:
@csadilek Thank you very much for the review!

For some reason the "reply comment" text box didn't appeared in one of your comments, so I will reply it here: about the `boolean root()` attribute inside `@WorkbenchPreference`, how would we infer it? I mean, even if it was an attribute if some other preference, the user could want another instance of it as a root preference in the tree view.

As we talked about on IRC, we will cover the dynamic plugins stuff in another PR.

csadilek:
@paulovmr thanks for the updates. 

Re: root = true, since, by default, preferences are at the root level now, how do you know what the root of a preference node is if root == false? In other words, how do you now what the parent node is in the UI for a preference with root == false? It seems to me that this should be know at compile time and then we don't need the root flag (since it defaults to true anyway). Happy to discuss on IRC. Not urgent and can be done later.

paulovmr:
@csadilek As we discussed on IRC, we will leave the flag for now, and think in a mecanism that replaces the flag and provides a way to a child insert itself into a parent.

ederign:
@paulovmr the root flag that @csadilek suggested will be fixed in a next PR right?

ederign:
@paulovmr  Great work, after you and @csadilek confirm this ^ , we can merge it to master!

paulovmr:
@ederign Yes, we agreed to change that in another PR.

PR #518, opened 2016-09-27
UF-386: Workbench Preferences Bean API - Default Value

paulovmr:


uberfire-extensions/uberfire-preferences/uberfire-preferences-backend/src/main/java/org/uberfire/ext/preferences/backend/ServerUsernameProvider.java
```diff
@@ -0,0 +1,47 @@
...
+
+import javax.enterprise.context.ContextNotActiveException;
+import javax.enterprise.context.Dependent;
+import javax.inject.Inject;
+
+import org.uberfire.ext.preferences.shared.UsernameProvider;
+import org.uberfire.rpc.SessionInfo;
+
+@Dependent
+public class ServerUsernameProvider implements UsernameProvider {
```
csadilek: Wouldn't it be enough to just inject SessionInfo? It returns a default user when not authenticated. What's the benefit of having this type?

csadilek:
+1 (one minor comment inline). Thanks!

uberfire-extensions/uberfire-preferences/uberfire-preferences-backend/src/main/java/org/uberfire/ext/preferences/backend/ServerUsernameProvider.java
paulovmr: The reason for this is that I needed to use this in a backend  `@Startup` bean. When the `SessionInfo`is injected there, it throws a `ContextNotActiveException` (because the `SessionInfo`'s producer is `@RequestScoped`, and no request was made). If I catch this exception directly in the class I was (a shared one), the GWT compilation would start to fail, presumably because it doesn't have this exception there. So that's why I had to make separated implementations for the server and the client.

ederign:
Thanks @paulovmr

uberfire-extensions/uberfire-preferences/uberfire-preferences-backend/src/main/java/org/uberfire/ext/preferences/backend/ServerUsernameProvider.java
csadilek: Right, but User isn't request scoped, so you could inject Instance<User>, no? It should have a default user when not authenticated....

csadilek: OK, I will take a look at that.

PR #523, opened 2016-10-04
Upgrade lienzo core to 2.0.277 and lienzo tests to 1.0.0-RC3.

romartin:
Upgrade lienzo artifacts to latest versions as done in latest kie parent POM. See https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/commit/2cfb9acadaf0630045870625044c5170dff01ca2

@manstis Here it is! :)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -52,8 +52,9 @@
     <uberfire.version>${project.version}</uberfire.version>
     <!-- Exception as it is needed for gwt-maven-plugin version-->
     <version.com.google.gwt>2.8.0-rc2</version.com.google.gwt>
-    <version.com.ahome-it.lienzo-core>2.0.270-RELEASE</version.com.ahome-it.lienzo-core>
-    <version.com.ahome-it.lienzo-tests>1.0.0-RC1</version.com.ahome-it.lienzo-tests>
+    <!-- when the new jboss-ip-bom 7.0.0.CR5 was released the lienzo versions can be removed -->
+    <version.com.ahome-it.lienzo-core>2.0.277-RELEASE</version.com.ahome-it.lienzo-core>
+    <version.com.ahome-it.lienzo-tests>1.0.0-RC3</version.com.ahome-it.lienzo-tests>
     <version.com.allen-sauer.gwt.dnd>3.3.3</version.com.allen-sauer.gwt.dnd>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -52,8 +52,8 @@
     <uberfire.version>${project.version}</uberfire.version>
     <!-- Exception as it is needed for gwt-maven-plugin version-->
     <version.com.google.gwt>2.8.0-rc2</version.com.google.gwt>
-    <version.com.ahome-it.lienzo-core>2.0.270-RELEASE</version.com.ahome-it.lienzo-core>
-    <version.com.ahome-it.lienzo-tests>1.0.0-RC1</version.com.ahome-it.lienzo-tests>
+    <version.com.ahome-it.lienzo-core>2.0.277-RELEASE</version.com.ahome-it.lienzo-core>
```
manstis: @romartin It'd probably be good to include the comment about being able to remove these versions when the next IP-BOM is released (and we change our parent ```pom``` too). See [here](https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/commit/8f7332a8a379eff8e8d28cf75df8696c951159ae). Furthermore do we not also need the supporting dependencies declared too? See [here](https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/commit/2cfb9acadaf0630045870625044c5170dff01ca2#diff-600376dffeb79835ede4a0b285078036R1648).

romartin:
Hi @manstis , just added a new commit with the comments you say, good point! About adding the supporting dependencies... wiell I dont see it really necessary unless some uf code depends on those, so then maven enforcer should complain... 

I had to add those in the kie parent pom as I'm using explicitly some of the lienzo nativetools classes on stunner, so maven enforcer rules were complaining. But in UF, if the build doesn't complain means that there are no concrete classes usage, so no really need to add those here, isnt it?  Just lemme know what you think, np for adding those as well if necessary. Thanks!

manstis:
@romartin Agreed (if the transient dependencies are not used in Uberfire, then no need to include them).

romartin:
@manstis yes would prefer adding those if really necessary... thanks for your consideration!!! :)

ederign:
+1 thanks @romartin

PR #528, opened 2016-10-07
[UF-416] missing filesystem unlock on Exception in JGitFileSystemProv…

tkobayas:
…ider.commit()


ederign:
@tkobayas great catch! Thanks for your contribution!

PR #535, opened 2016-10-20
[0.9.x] RHBRMS-2519: Upload File doesn't support multi-byte symbols in file name

karreiro:
Only on EAP 6.4 (running the version from the 6.5.x branch), [this solution](https://github.com/droolsjbpm/kie-docker-ci-images/pull/7) was not enough to solve this issue: https://issues.jboss.org/browse/RHBRMS-2519

So this workaround is needed:
```
request.setCharacterEncoding( "UTF-8" );
```

### DIFF
--- a/uberfire-server/src/main/java/org/uberfire/server/FileUploadServlet.java
+++ b/uberfire-server/src/main/java/org/uberfire/server/FileUploadServlet.java
@@ -50,6 +50,8 @@ public class FileUploadServlet
                            HttpServletResponse response ) throws ServletException, IOException {
 
         try {
+            request.setCharacterEncoding( "UTF-8" );
+
             if ( request.getParameter( PARAM_PATH ) != null ) {
 
                 //See https://bugzilla.redhat.com/show_bug.cgi?id=1202926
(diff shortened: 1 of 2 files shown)


karreiro:
@ederign Could you please review this? :-)

manstis:
@karreiro Have a look at ```FileUploadFormEncoder``` it could be related - and if it is not used consistently it could explain why some Editors are reported as working OK and others not.

karreiro:
@manstis Thanks for the clue :-) However, I think that the file uploader form is already using the `FileUploadFormEncoder` see [here](https://github.com/uberfire/uberfire-extensions/blob/0.9.x/uberfire-widgets/uberfire-widgets-core/uberfire-widgets-core-client/src/main/java/org/uberfire/ext/widgets/core/client/editors/defaulteditor/DefaultEditorFileUploadBase.java#L74).

ederign:
@karreiro why don't put that also on master?

karreiro:
@ederign because this bug *only* happens on EAP 6.4. The master is fine. :-)

Rikkola:
So looks like this is fixed in the newer versions of EAP ( and probably in other container ). What we are doing here is duct tape since we can not fix EAP.
I'm guessing this does no harm so +1.

ederign:
thanks @karreiro

PR #536, opened 2016-10-21
Fix for drools-wb

paulovmr:


manstis:
I'd like to understand how ```@Perspective``` can be called before ```@OnStartup```. The fix ensures the ```perspective``` is initialised but I don't understand how a request for the perspective definition could be made without "going to" the perspective (i.e. ```@OnStartup``` called). The PR is fine BTW.. just want to understand the reason why it is needed.

paulovmr:
Hi @manstis , the call to `@Perspective` is made when the [drools-wb-webapp's menu is built](https://github.com/droolsjbpm/kie-wb-common/blob/master/kie-wb-common-screens/kie-wb-common-workbench/kie-wb-common-workbench-client/src/main/java/org/kie/workbench/common/workbench/client/menu/DefaultWorkbenchFeaturesMenusHelper.java#L214), because the method `getDefaultPerspectiveLayout()` is called so the perspective name can be obtained for sorting purposes.

uberfire-extensions/uberfire-preferences/uberfire-preferences-client/src/main/java/org/uberfire/ext/preferences/client/central/PreferencesCentralPerspective.java
```diff
@@ -77,6 +77,13 @@ public void perspectiveChangeEvent( @Observes UberfireDockReadyEvent dockReadyEv
 
     @Perspective
     public PerspectiveDefinition getPerspective() {
+        if ( perspective == null ) {
+            PerspectiveDefinition perspective = new PerspectiveDefinitionImpl( MultiListWorkbenchPanelPresenter.class.getName() );
+            perspective.setName( "Preferences" );
```
manstis: Should this be i18n'ed?

uberfire-extensions/uberfire-preferences/uberfire-preferences-client/src/main/java/org/uberfire/ext/preferences/client/central/PreferencesCentralPerspective.java
```diff
@@ -77,6 +77,13 @@ public void perspectiveChangeEvent( @Observes UberfireDockReadyEvent dockReadyEv
 
     @Perspective
     public PerspectiveDefinition getPerspective() {
+        if ( perspective == null ) {
```
manstis: Should we have a test?

uberfire-extensions/uberfire-preferences/uberfire-preferences-client/src/main/java/org/uberfire/ext/preferences/client/central/PreferencesCentralPerspective.java
paulovmr: No, this name is not displayed anywhere.

ederign:
thanks @paulovmr

csadilek:
@ederign @paulovmr please make sure your commit messages make sense from an UF perspective e.g. "Fix to ensure perspectives are initialized (details below)" is better than "Fix drools-wb". What is drools-wb to an UF user? ;)

PR #539, opened 2016-10-24
UF-424: return empty collection for groups and roles

sutaakar:


ederign:
@sutaakar thanks for your contribution. 

@romartin could you please take a look on this one?

romartin:
+1
Thanks @sutaakar , @ederign !

PR #541, opened 2016-10-24
UF-320: Adding counters for roles, groups and users on admin page

paulovmr:
* Handled admin page wrong usage errors.
* Fixed admin page container height.
* Added counters for roles, groups and users on admin page on uberfire-security-management-webapp.

adminpage2


csadilek:
+1

PR #548, opened 2016-11-02
GUVNOR-2736: enables the re-use of the applied response security  hea…

porcelli:
…ders


csadilek:
+1

csadilek:
See also: https://github.com/droolsjbpm/kie-wb-distributions/pull/395

PR #553, opened 2016-11-09
[UF-370]: update security modal dialogs layout

tomasdavidorg:
Moves primary buttons to left to be the first button as in other modal dialogs and moves buttons from center to right in change password dialog.


ederign:
Thanks @tomasdavidorg

PR #557, opened 2016-11-11
GUVNOR-2734: Guided Decision Table Editor: Close Analysis Report for graph editor

manstis:
See https://issues.jboss.org/browse/GUVNOR-2734

Cached "Connectors" were not being cleared when the Canvas was cleared.


manstis:
@Rikkola This fixes the issues you were having [here](https://github.com/Rikkola/drools-wb/commits/GUVNOR-2734-open-close-vv-reporting).

Rikkola:
Thank you

ederign:
Thanks @manstis

PR #561, opened 2016-11-13
Simplified API for Uberfire Breadcrumbs and also new breadcrumbs area:toolbar

ederign:


ederign:
Parts of the ensemble:

https://github.com/uberfire/uberfire/pull/561
https://github.com/droolsjbpm/kie-wb-common/pull/561
https://github.com/droolsjbpm/drools-wb/pull/339
https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/314
https://github.com/droolsjbpm/kie-wb-distributions/pull/405

porcelli:
+1, but with the following comments:

This is Ok for this interaction, but for next interation I'd like to see some changes:

 - Breadcrumbs should be a Header, like Brand and Navigation (menus). We already covered this with ordered feature of the Headers.
 - Maybe add a generic Header that would enable different content (ie. remove the Bradcrumb add toolbar capability). I think have directly into Breadcrumb the addition of a toolbar is mixing concerns. Breadcrumb should be just a navigation capability.
 - Have the current breadcrumb mechanism always available looks weird where it's not used.

ederign:
@porcelli thanks. I'll do that in the next iteration,.

ederign:
This is the right on kie-wb-common: https://github.com/droolsjbpm/kie-wb-common/pull/567

PR #565, opened 2016-11-19
RHBRMS-2687: [Screen Plugin] Fail to upload picture on WAS and IE11

karreiro:
See https://issues.jboss.org/browse/RHBRMS-2687
____

- Before:
a2

- After:
a1


karreiro:
@etirelli @ederign Could you please take a look?

ederign:
Hi @karreiro thanks for the fix. Did u test also in *nix right?

karreiro:
Hi @ederign Yes, I did :-)

PR #567, opened 2016-11-22
GUVNOR-2750: Guided Decision Table Editor: Multi-line Selected Cell highlight is not clipped vertically with Header

manstis:
See https://issues.jboss.org/browse/GUVNOR-2750


manstis:
@Rikkola @karreiro Could you please review when you get a chance.

ederign:
thanks @manstis

PR #574, opened 2016-11-30
[UF-444] Avoid concurrent Freemarker setup

psiroky:
The annotation processor sometimes failed when used
concurrently (e.g. in parallel Maven build). The reason
seems to be the fact that Freemarker's Configuration class
(or its dependencies) is not thread safe.

Synchronizing the Freemarker setup will likely make the init
sligthly slower, but since this is only build-time
(compile-time) processor it should not matter much.

@ederign, @manstis this seems to fix the issue we talked about today. I am running more build locally to make sure. However, I am not 100% sure about the implications of this change. WDYT? Is this OK?


manstis:
@psiroky Heh.. implications.. heh! Well, out code generators were never written to work in a multi-threaded environment, so IMO anything that fixes any issues is good.. so +1 from me (and looks better than synchronizing the whole ```generate(..)``` method).

psiroky:
30 builds of kie-wb-common were successful in a row (running parallel build with 10 threads), so it indeed seems to fix the issue.

PR #575, opened 2016-12-01
UF-445 : Move UUID from uberfire-wires-core-api to uberfire-commons

Rikkola:
https://issues.jboss.org/browse/UF-445

Depends on this:
https://github.com/droolsjbpm/drools-wb/pull/354

### DIFF
--- a/uberfire-commons/src/main/java/org/uberfire/commons/uuid/UUID.java
+++ b/uberfire-commons/src/main/java/org/uberfire/commons/uuid/UUID.java
@@ -20,10 +20,15 @@
  * and open the template in the editor.
  */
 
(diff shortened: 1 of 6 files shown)


manstis:
Do ```droolsjbpm``` repositories need any change following this (other than ```drools-wb``` to re-use this!)

Rikkola:
Spent so much time in my local web worker branch that I forgot drools-wb already used this, but looks like that is the only one. Wires is just used by drools-wb, checked with Maven.

ederign:
@Rikkola we are good to go? Or do we need a syncronized merge across other repos?

ederign:
soooo @porcelli  ?

ederign:
thanks @Rikkola

PR #576, opened 2016-12-01
Properly depend on freemarker + fix compiler args

psiroky:


ederign:
@psiroky did you try this on eclipse? Because on the old commit message, jfuerth did this in order to fix a Eclipse issue. We need to make sure that uberfire still works in eclipse env.

psiroky:
@ederign I did not. Using shading to fix some Eclipse issue is just horrible. I'll try to open the project in eclipse to see what happens.

psiroky:
I am not an Eclipse user, so not sure what to look for (and the old commit message does not say either). The project opens fine and I can execute tests in `uberfire-workbench-processors-tests` module.

ederign:
+1 thanks @psiroky

PR #578, opened 2016-12-03
UF-444: Avoid concurrent Freemarker setup (uberfire-preferences-processors)

paulovmr:
@psiroky Could you please take a look? As mentioned in my email, it would be nice to test it in a Jenkins/release build, checking if the "Home" > "Admin" page is loading correctly.

### DIFF
--- a/uberfire-extensions/uberfire-preferences/uberfire-preferences-processors/src/main/java/org/uberfire/ext/preferences/processors/AbstractGenerator.java
+++ b/uberfire-extensions/uberfire-preferences/uberfire-preferences-processors/src/main/java/org/uberfire/ext/preferences/processors/AbstractGenerator.java
@@ -43,10 +43,12 @@ public abstract class AbstractGenerator {
             throw new NoClassDefFoundError( "Failing for testing purposes" );
         }
         try {
-            config = new Configuration();
-            config.setClassForTemplateLoading( getClass(),
-                                               "templates" );
-            config.setObjectWrapper( new DefaultObjectWrapper() );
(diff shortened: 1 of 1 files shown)


manstis:
There are Arquilian tests in kie-wb-distributions/kie-wb-tests/kie-wb-tests-gui that check various things including login, goto authoring/library, create project, build and deploy etc. You could write another test for preferences.

psiroky:
@paulovmr could we somehow refactor the code to avoid the duplication? I've noticed this particular piece of code was copy pasted into kie-wb-common as well: https://github.com/droolsjbpm/kie-wb-common/blob/master/kie-wb-common-stunner/kie-wb-common-stunner-core/kie-wb-common-stunner-processors/src/main/java/org/kie/workbench/common/stunner/core/processors/AbstractAdapterGenerator.java#L40

The issues is not that easily reproducible I am afraid. Once merged we can run couple of builds to see what happens. With forking enabled (https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/commit/60ddedd7046f8ffdc9563c642a6e99bca85c509d) the issue is for now gone, so we would also need to disable the forking again. I am actually looking at that, but I am still not able to reproduce this easily on a local machine.

paulovmr:
jenkins retest this

paulovmr:
@manstis Test created and merged: https://github.com/droolsjbpm/kie-wb-distributions/pull/429

@psiroky Now that we are removing the shading from freemarker dependency, the refactor will be possible. I will close this PR and soon open another one with the refactor included.

PR #581, opened 2016-12-09
[UF-448] Upgrade Templated Perspectives to Errai UI

ederign:


ederign:
Please don't merge this yet. I'm still upgrading a ton of showcases.

ederign:
https://github.com/dashbuilder/dashbuilder/pull/264

ederign:
jenkins retest this

uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/util/ErraiTemplateUtil.java
```diff
@@ -0,0 +1,94 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.uberfire.client.util;
+
+import org.jboss.errai.common.client.dom.Event;
+import org.jboss.errai.common.client.dom.HTMLElement;
+import org.jboss.errai.common.client.ui.ElementWrapperWidget;
+
+public class ErraiTemplateUtil {
```
porcelli: @mbarkley @csadilek @ederign what about move this to Errai? (except the UF_PERSPECTIVE_COMPONENT of course)

uberfire-workbench/uberfire-workbench-client/src/main/resources/org/uberfire/client/resources/css/workbench.css
```diff
@@ -161,4 +161,63 @@
...
+.uf-perspective-container{
+    height: 100%;
+}
+
+.uf-perspective-row-12{
+    height: 100%;
+}
+
+.uf-perspective-row-11{
+    height:91.66666667%;
```
porcelli: what about use calc(100%  / n) ? this is supported for all major browsers as you can see http://caniuse.com/#feat=calc

uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/util/ErraiTemplateUtil.java
ederign: Yes,  if they both think that will be good for Errai, I can create a new PR for ERRAi and remove this code from here.

uberfire-workbench/uberfire-workbench-client/src/main/resources/org/uberfire/client/resources/css/workbench.css
ederign: My understanding of calc is that is use case for dynamic stuff (i.e. calc(100%-3px).

uberfire-workbench/uberfire-workbench-client/src/main/resources/org/uberfire/client/resources/css/workbench.css
```diff
@@ -161,4 +161,63 @@
     -moz-user-select: none;
     -ms-user-select: none;
     user-select: none;
+}
+
+.uf-perspective-container{
```
karreiro: @ederign There are some formatting issues in this file. You could change the declarations from `.something{` to `.something {`.

ederign:
@karreiro fixed! Thanks.

ederign:
jenkins retest this

uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/util/ErraiTemplateUtil.java
```diff
@@ -0,0 +1,94 @@
...
+                        newClass += " ";
+                    }
+                    newClass += aClass;
+                }
+            }
+            element.setClassName( newClass );
+        }
+    }
+
+    public static boolean hasClassName( HTMLElement element, String cssClass ) {
```
mbarkley: There is a method for this in Errai's [DOMUtil](https://github.com/errai/errai/blob/master/errai-common/src/main/java/org/jboss/errai/common/client/dom/DOMUtil.java) class.

uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/util/ErraiTemplateUtil.java
```diff
@@ -0,0 +1,94 @@
...
+            }
+        }
+        return false;
+    }
+
+    public static native String extractOffSetWidth( HTMLElement el ) /*-{
+        return el.offsetWidth;
+    }-*/;
+
+    public static native String extractWidth( HTMLElement el ) /*-{
```
mbarkley: I also think this would be better as a JS interop method on HTMLElement. That being said, this will only give you a meaningful value if the width of an element has been explicitly assigned. It's usually better to get the computed width of an element, I think.

uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/util/ErraiTemplateUtil.java
```diff
@@ -0,0 +1,94 @@
...
+        return el.width;
+    }-*/;
+
+
+    public static native String extractClientY( Event e ) /*-{
```
mbarkley: There are JS interop methods on [MouseEvent](https://github.com/errai/errai/blob/master/errai-common/src/main/java/org/jboss/errai/common/client/dom/MouseEvent.java) for this and the method below.

uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/util/ErraiTemplateUtil.java
```diff
@@ -0,0 +1,94 @@
...
+        return e.clientY;
+    }-*/;
+
+    public static native String extractClientX( Event e ) /*-{
+        return e.clientX;
+    }-*/;
+
+    public static int extractAbsoluteTop( HTMLElement el ) {
```
mbarkley: This method and the one below would probably make good additions to Errai's DOMUtil class. That being said, it would be better to use a JSNI method to simply cast the HTMLElement to a GWT element. This implementation has  memory leak for HTMLElements that do not already have ElementWrapperWidgets.

uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/util/ErraiTemplateUtil.java
mbarkley: @ederign I've added comments inline. A couple of the methods could be moved to Errai's DOMUtil class. Most of the others ought to be turned into JS interop methods on the appropriate types.

ederign:
Depends on https://github.com/errai/errai/pull/214

ederign:
@porcelli the errai PR is merged. Do we need something else?

ederign:
jenkins retest this

PR #584, opened 2016-12-19
Add missing license headers

csadilek:
@manstis adding a lot of missing license headers in various UF classes.

/cc @ederign


csadilek:
jenkins retest this.

PR #586, opened 2016-12-21
Avoid logging of plugin watcher errors during shutdown

csadilek:
- An additional check makes sure the watcher is active
before processing plugin changes. This makes it much
less likely to happen but there's still a time window
in which it could occur

- Therefore if an error is thrown during shutdown
we just log it in debug level


csadilek:
@manstis as discussed, could you please take a look? thanks.

csadilek:
Test failure unrelated. Jira already created for that one. Jenkins retest this.

PR #587, opened 2016-12-22
UF-452: if in cluster, user won't see any error on bus is offline. If not in cluster, it will show error, but not force refresh

porcelli:


uberfire-workbench/uberfire-workbench-client-backend/src/main/java/org/uberfire/client/WorkbenchBackendEntryPoint.java
```diff
@@ -19,54 +19,76 @@
...
-    private void startApp() {
-        bus.addTransportErrorHandler( new TransportErrorHandler() {
+    @PostConstruct
+    public void postConstruct() {
+        bus.addLifecycleListener( new BusLifecycleAdapter() {
             @Override
-            public void onError( TransportError error ) {
-                if ( askRefresh ) {
+            public void busOnline( final BusLifecycleEvent e ) {
+                logger.error( "Bus is back online." );
```
csadilek: logger.info.

csadilek:
0.9.x: f1127db55c3407587f927b4c1ed7ba5aac99d277

PR #588, opened 2016-12-22
UF-452: if in cluster, user won't see any error on bus is offline. If not in cluster, it will show error, but not force refresh

porcelli:


uberfire-workbench/uberfire-workbench-client-backend/src/main/java/org/uberfire/client/WorkbenchBackendEntryPoint.java
```diff
@@ -19,54 +19,76 @@
...
-    private void startApp() {
-        bus.addTransportErrorHandler( new TransportErrorHandler() {
+    @PostConstruct
+    public void postConstruct() {
+        bus.addLifecycleListener( new BusLifecycleAdapter() {
             @Override
-            public void onError( TransportError error ) {
-                if ( askRefresh ) {
+            public void busOnline( final BusLifecycleEvent e ) {
+                logger.error( "Bus is back online." );
```
csadilek: This should be logger.info

csadilek:
@porcelli I made the last remaining change for you directly. So, you don't have to get back to this .

master: 6df0a174cce1a0388f01640a8c0436bdabaecd32

porcelli:
@csadilek thx!

manstis:
BTW there was a missing license header in the test.

porcelli:
@manstis thx, i'll update in a few.

PR #601, opened 2017-01-10
Fix initialization ordering problem in GridLayerRedrawManager

csadilek:
The comparator was initialized after the singleton instance was created,
causing a null comparator to be passed to the constructor of the commands
TreeSet and hence natural ordering to be used instead which didn't work
as Commands are not Comparable.

Since this type is a singleton it shouldn't have any other static fields
than instance to guarantee that the instance is in a proper state when
created.

This also caused ClassCastExceptions with generateJsInteropExports=true.

### DIFF
--- a/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/layer/impl/GridLayerRedrawManager.java
+++ b/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/layer/impl/GridLayerRedrawManager.java
@@ -28,7 +28,7 @@ public class GridLayerRedrawManager {
 
     private static final GridLayerRedrawManager instance = new GridLayerRedrawManager();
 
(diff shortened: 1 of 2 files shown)


csadilek:
@manstis @romartin @ederign 

Can you please take a look and merge, thanks. We ran into this when integrating GDT in Eclipse. It was pretty hard to track down as all we got were cryptic ClassCastException. The problem is that the provided comparator actually isn't being used as it's initialized after the singleton instance and therefore null when provided to the commands TreeSet constructor. This causes natural ordering to be used instead of the comparator and all kinds of problems since Commands are not comparable. In our case, it was ClassCastExceptions when a command got cast to comparable against its will :).

manstis:
Looks fine; but I'll check a few things tomorrow (wires showcase depends on this class for the grids demo). I have found class instantiation and member initialisation is "unexpected" in GWT generated Java Script vs Java in the past..

csadilek:
@manstis ok, the ordering definitely was incorrect, in both Java and JavaScript hence the removal of the incorrect static identifier. With both the instance and the Comparator being static neither the jvm nor the javascript engine can now what to do first ;). Thanks for giving it a spin tomorrow.

manstis:
@csadilek @ederign Ran wires-webapp with this change locally; and everything is hunky-dory.

PR #611, opened 2017-01-24
[UF-453]: WorkspaceScoped Beans added

adrielparedes:
_NOTE: I don't know if the files from the beginning should be there, so please let me know._

The Idea is to have a CDI Annotation to replace the @ApplicationScoped with it. This annotations lets a bean enter a Workspace context, so those beans have a single instance per workspace. Right now there is only one workspace per user, but that is going to change in the future.

### DIFF
--- a/uberfire-backend/uberfire-backend-cdi/src/main/java/org/uberfire/backend/server/cdi/workspace/WorkspaceManager.java
+++ b/uberfire-backend/uberfire-backend-cdi/src/main/java/org/uberfire/backend/server/cdi/workspace/WorkspaceManager.java
@@ -1,0 +1,152 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.uberfire.backend.server.cdi.workspace;
+
+import java.util.NoSuchElementException;
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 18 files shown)


uberfire-backend/uberfire-backend-cdi/src/main/java/org/uberfire/backend/server/cdi/workspace/WorkspaceManager.java
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    @Inject
+    public WorkspaceManager( WorkspaceManagerPreferences workspaceManagerPreferences ) {
+        this.preferences = workspaceManagerPreferences;
+    }
+
+    @PostConstruct
+    public void initialize() {
+        this.workspaces = new HashMap<>();
```
csadilek: Can we use a ConcurrentHashMap instead? Then we don't have to synchronize all get/put methods. I mean ```createCache``` would likely still need to by synchronized, but not getOrCreateWorkSpace and getBean, putBean.

uberfire-backend/uberfire-backend-cdi/src/main/java/org/uberfire/backend/server/cdi/workspace/WorkspaceScoped.java
```diff
@@ -0,0 +1,36 @@
...
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import javax.enterprise.context.NormalScope;
+import javax.interceptor.InterceptorBinding;
+
```
csadilek: Can we get JavaDocs for this please, similar to AppScoped etc. to describe the scope?

uberfire-backend/uberfire-backend-cdi/src/main/java/org/uberfire/backend/server/cdi/workspace/WorkspaceManager.java
adrielparedes: @csadilek good point, I will add them now

uberfire-backend/uberfire-backend-cdi/src/main/java/org/uberfire/backend/server/cdi/workspace/WorkspaceScoped.java
adrielparedes: @csadilek sure, I tried not to forget anything, but can fail :p

adrielparedes:
jenkins retest this

PR #616, opened 2017-01-26
Intermittent test failures of WildflyGroupsPropertiesManagerTest

ederign:
I hope that fix the issue 🤞. It just a hope because I cannot reproduce it locally.

But seems promising:

testDeleteGroup fails because a ConcurrentModificationException probably on 

````
@Override
    public void delete(String... identifiers) throws SecurityManagementException {
        if (identifiers == null) throw new NullPointerException();
        try {
            Set<Map.Entry<Object, Object>> propertiesSet = groupsPropertiesFileLoader.getProperties().entrySet();
            if (!propertiesSet.isEmpty()) {
                for (Map.Entry<Object, Object> entry : propertiesSet) {
                  ...
                }
            }
        } catch (Exception e) {
            LOG.error("Error removing the folowing group names: " + identifiers, e);
            throw new SecurityManagementException(e);
        }
    }
````
 and this is also modified by:
````
public class WildflyGroupPropertiesManager { ...
   @Override
    public void initialize(UserSystemManager userSystemManager) throws Exception {
        this.groupsPropertiesFileLoader = getFileLoader(getGroupsFilePath());
    }
````
called here:

````
 @Before
    public void setup() throws Exception {
        ...
        groupsPropertiesManager.initialize(userSystemManager);
    }
````


ederign:
doesn't work ;)

ederign:
investigating...

ederign:
jenkins retest this

ederign:
jenkins retest this

ederign:
jenkins retest this

PR #619, opened 2017-02-01
Fix support for menus in dynamic plugins

csadilek:
- Make sure EnableStateChangeListeners can be shared
across scripts including their enclosing scope

- Make sure menu items are iterable even if provided
by external script


csadilek:
@ederign This fixes both issues we were debugging today. The state change listener should have always been native, but since the listeners weren't called when last tested it didn't matter.

Feel free to close and merge with your other refactorings or as is.

ederign:
Awesome @csadilek! 👍 

screenshot 2017-02-02 08 54 12

PR #621, opened 2017-02-02
Refactor EnableStateChangeListener to a JsFunction

csadilek:
- This allows us to postpone activating jsinteropexports for all apps
while at the same time preventing the compiler optimization (inlining)
that caused problems for dynamic menus


csadilek:
@manstis @ederign this seems to be the compromise to get rid of the compiler optimization (method inlining) when compiling the plugins, and also not force us to enable jsinteropexports everywhere right away.

csadilek:
@manstis I cleared all caches and tried with drools-wb and uberfire (with and without problems). All these cases work fine for me. /cc @ederign

csadilek:
Ah testDeleteGroup our old friend, jenkins retest this.

csadilek:
jenkins restest this.

PR #623, opened 2017-02-03
For discussion.. signalling Indexer tests in lieu of Thread.sleep(...)

manstis:
@ederign and I think @psiroky spoke about the obvious "elephant" in the room... indexing tests that like to wait. Thanks to @porcelli original code and the virtues of copy and paste we now have a lot of indexer tests across lots of repositories that happily wait. This PR includes a potential fix - or basis for a fix - for the wait state.

The code that runs the indexers consists of two nested threads; one runs constantly (or at least until a VFS is disposed) and it's nested child in reaction to a set of ```WatchEvents``` raised from VFS. It's the inner child that calls the indexers. 

My change reports completion of the inner worker thread to an observer that can then perform tasks at different steps in a tests sequence. The test (or author) needs to know how many times the inner thread is invoked as you need to know the stage when the indexes should contain the correct information; for example ```ioService.delete(..)``` leads to the inner thread being started twice - one for the "real file" and again for "dot file" whereas ```ioService.write(..)``` leads to the inner thread being started once - for both the "real file" and "dot file```. So the tests currently set up "checks" to perform per completion of the inner thread. Some checks are NOPs as I don't care what the index status was at the time. 

Anyway, not for merge, but to start discussion about how to improve our indexing tests..


manstis:
@csadilek @psiroky @ederign @porcelli ... for opinions please.

manstis:
Jenkins please retest this.

porcelli:
@manstis what about, instead change our execution model (adding the observer), why not use Byteman to block/release/check? We already have some Byteman tests in our codebase: https://github.com/uberfire/uberfire/blob/master/uberfire-nio2-backport/uberfire-nio2-impls/uberfire-nio2-jgit/src/test/java/org/uberfire/java/nio/fs/jgit/JGitFileSystemProviderBytemanTest.java

csadilek:
I like this solution as it's low on complexity and makes the tests easy to read (not that I dislike the Byteman approach, but it is significantly more complex).

One minor suggestion I have is to improve the API used in tests...to add a call similar to this:
```
observer.onEventCount(3, () -> assertEvents());
```
instead of
```
observer.addInformationCallback(IndexObserverCallback.NOP);
observer.addInformationCallback(IndexObserverCallback.NOP);
observer.addInformationCallback(() -> assertEvents(3));
```

manstis:
I guess my PR reflects the difference between me and Alexandre... I am simple (in mind) whereas he is complex. Let's wait for more comments (in any) before making any decision. Thanks for the feedback (so far).

psiroky:
I am also in favor of this approach when compared to Byteman.

ederign:
That solution could be also be applied to fix this intermittent failing test https://kie-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/uberfire-pullrequests/794/testReport/junit/org.uberfire.ext.metadata.io/BatchIndexConcurrenyTest/testSingleConcurrentBatchIndexExecution/

manstis:
Jenkins please retest this.

manstis:
Jenkins please retest this

manstis:
Jenkins please retest this

ederign:
@porcelli are you working on a POC for byteman approach?

porcelli:
PR #628 as byteman alternative

PR #624, opened 2017-02-03
Flat Web Jar Uberfire pom.xml Support

ederign:
Dynamic Plugins now generates web jar files in maven build.

**Jar Structure:**
screenshot 2017-02-03 17 48 00

**File Size:**

screenshot 2017-02-03 17 45 11

### DIFF
--- a/uberfire-showcase/uberfire-dynamic-plugin/pom.xml
+++ b/uberfire-showcase/uberfire-dynamic-plugin/pom.xml
@@ -32,6 +32,10 @@
   <name>UberFire Dynamic Plugin</name>
   <description>UberFire Dynamic Plugin</description>
 
+  <properties>
+    <gwt.module.name>org.uberfire.UberfireDynamicPlugin</gwt.module.name>
+  </properties>
+
   <dependencies>
     <dependency>
       <groupId>org.uberfire</groupId>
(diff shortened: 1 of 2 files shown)


ederign:
Updated version after @csadilek requests:
Get hide of JS dir and assembly in the jar
https://gist.github.com/ederign/65742312bfee32947d069d9a82fc2c33

Make sure that it works with classifier:

screenshot 2017-02-06 13 09 38

ederign:
screenshot 2017-02-06 15 10 56

Found it. That was happening because uberfire-dynamic-plugin-1.0.0-SNAPSHOT-jar-with-dependencies.jar contains references to all transitive dependencies. If we place them as "provided" https://gist.github.com/anonymous/9e95ec35b56c7b3db24259c32f82a473 we will get our clean fat jar.

Question is: Should we do this in all future web jar pom's? Should I add this to this PR?

csadilek:
@ederign I think this is a home-made "problem" as we want to allow for projects being both dynamic plugins and a static GWT modules. So, I think it's fine to set all dependencies to provided if  the goal is to just build a dynamic plugin. In most cases these projects will just be webjars anyway. If someone wants to build a webjar and a static module at the same time (using a single pom) then they're going to have live with the dependencies being brought in.

psiroky:
Yeah, this is the default Maven behavior. It will include the transitive dependencies for all the resulting artifacts, because they both share the same pom.xml.

PR #627, opened 2017-02-06
RHBPMS-1889: Search by Find form does not work.

manstis:
Git internal "dot" files were being converted to a non-"dot" form which could then not be found on the underlying VFS. This fix skips indexing of git internal "dot" files.


manstis:
Jenkins please retest this.

PR #628, opened 2017-02-08
For discussion.. proposal that not only removes Thread.sleep but also remove the need of the Observer type

porcelli:
This PR is an alternative for #623


porcelli:
@ederign @csadilek @manstis @psiroky please

manstis:
@porcelli Great, does the same thing without touching _our_ source code. 

A couple of comments:

1. The parameter for ```waitForCountDown( int )``` doesn't appear to do anything. I assume we can still have JUnit "timeout" tests should we have an error somewhere (and I assume Byteman, in the event of a timeout would remove it's instrumented code)
2. Can we remove the need for "test specific" scripts? (there may be rare cases developers can provide more sophisticated scripts, but for our general "does the indexer work" type test having the ```count down```, ```return startBatch value```reset counter``` rules are enough).
3. Rule ```reset counter``` should probably reset _all_ counters? (you've added support for more fine grained observation in your ```index.btm``` file, although we're only using "index commits" in tests at the moment).

Otherwise I'm completely in favour of it.

porcelli:
@ederign @csadilek  @psiroky  most issues pointed by @manstis  already fixed

porcelli:
jenkins please retest this

porcelli:
(btw: failing test doesn't seem to be related)

uberfire-extensions/uberfire-metadata/uberfire-metadata-commons-io/src/test/java/org/uberfire/ext/metadata/io/IOServiceIndexedDeleteFileTest.java
```diff
@@ -23,73 +23,84 @@
...
 public class IOServiceIndexedDeleteFileTest extends BaseIndexTest {
 
     @Override
     protected String[] getRepositoryNames() {
-        return new String[]{this.getClass().getSimpleName()};
+        return new String[]{ this.getClass().getSimpleName() };
     }
 
     @Test
+    @BMScript(value = "byteman/first_time.btm")
```
manstis: IDK if exposing "reset" (for the Byteman counters) and using ```setupCountDown( 1 )``` would be more obvious than the purpose of this script (which ensures the first write doesn't block; after which you use ```setupCountDown(..)```).. Just a thought. Then we wouldn't need _any_ test-specific scripts.

uberfire-extensions/uberfire-metadata/uberfire-metadata-commons-io/src/test/java/org/uberfire/ext/metadata/io/IOServiceIndexedDotFileGitInternalImplTest.java
```diff
@@ -44,6 +48,7 @@
     }
 
     @Test
+    @BMScript(value = "byteman/first_time.btm")
```
manstis: Would ```setupCountDown( 1 )``` not be preferential to using "first time"?

uberfire-extensions/uberfire-metadata/uberfire-metadata-commons-io/src/test/java/org/uberfire/ext/metadata/io/IOServiceIndexedDeleteFileTest.java
porcelli: the problem is that if you just expect it execute once.. you'd need to use setupCountDown( 0 )... what is the start state... and it would fail

porcelli:
@manstis no more "first_time"

manstis:
Jenkins please retest this.

manstis:
@porcelli Gets +1 from me (same principle as https://github.com/uberfire/uberfire/pull/623 but better in may ways).

porcelli:
seems that the test on WildFly is really broken, but it's not related to this PR
cc/ @romartin

manstis:
@porcelli See https://github.com/uberfire/uberfire/pull/629

manstis:
Jenkins please retest this.

porcelli:
@csadilek @psiroky any comments?

psiroky:
Not really :) It removes the Thread.sleeps() so that's great!

PR #629, opened 2017-02-08
Wildfly Security Management provider - Bug fixing

romartin:
Hey @ederign @csadilek @psiroky 

Took a look at the  race condition that sometimes is causing a concurrent a modification exceptio.  The issue I think is just due to the library implementation for that race condition and how I was using it.

If you look [here](https://gist.github.com/anonymous/fc3a5c6cdd7b05a08344c6036e32e9b9#file-propertiesfileloader-java-L172), it just says that you can have different threads accessing that resrouce, but it's not really an implemntation to serously consider for ensuring complete thread-safe support... Anyway, as far as I see they use to synchronize the PropertiesFileLoader instance itsel on the `getProperties()` method, so it was a bad idea to keep a reference for that properties in the uberfire wildfly groups manager and then iterating... because any other thread, even from the users manager (both users/roles files are close related), could cause this concurrent exception.

Summarizing, I have done a review and this seems the only point where diong that kind of iteration and update... so I think this fix could be enought. I will force J  to rebuild this PR several times and will see... 

Lemme know your thioughts! Thanks!

### DIFF
--- a/uberfire-extensions/uberfire-security/uberfire-security-management/uberfire-security-management-wildfly/src/main/java/org/uberfire/ext/security/management/wildfly/properties/WildflyGroupPropertiesManager.java
+++ b/uberfire-extensions/uberfire-security/uberfire-security-management/uberfire-security-management-wildfly/src/main/java/org/uberfire/ext/security/management/wildfly/properties/WildflyGroupPropertiesManager.java
@@ -21,7 +21,6 @@ import org.jboss.as.domain.management.security.PropertiesFileLoader;
 import org.jboss.errai.security.shared.api.Group;
 import org.jboss.errai.security.shared.api.GroupImpl;
 import org.jboss.errai.security.shared.api.Role;
-import org.jboss.errai.security.shared.api.identity.User;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.uberfire.commons.config.ConfigProperties;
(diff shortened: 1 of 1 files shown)


romartin:
basically it's just relying on calling the `getProperties()` on each iteration, which is being synchronized... (trying to explain better :))

romartin:
jenkins retest this

romartin:
jenkins retest this

romartin:
jenkins retest this

csadilek:
@romartin thanks, seems good to me. I would move the getProperties() call outside the forEach iteration though.

romartin:
hey @csadilek , not sure if I'm following you sorry but what you say I think it's more or less what was done before, so I mean the fix is about calling that getProperties on each loop iteration, as this getter ist he one that gets syncrhonized... if I keep a copy the properties outside the loop and they're being updated, I would not realize if some property I want to change has been probably deleted by other client... does it makes sense for you? thanks!

romartin:
so at each loop iteration, calling the `getProperties()`, ensures that I have the latest stuff... as this methods does that logic

romartin:
well you probably are not in scope sorry, I say  this because the implementastion of the `getProperties()` method does "loadAsRequired" as first operation, so it ensures that properties are the latest ones... and this is the syncrhonized stuff

romartin:
BTW fired several jenkins builds and all are success

csadilek:
@romartin yes, I think this fixes the issue, but what is unclear to me is this: Within your loop who/what else is changing these properties "concurrently"? The calls to deleteGroupsFromSerliazedValue and updadeGroupProperty in the same loop body?

romartin:
@csadilek well good point, lemme explain - That propertiesFileLoader instance, due to its implementation and the way it does syncrhonization, on a container context (cdi) is being used as a member for an application scoped bean, so different thread can be accessing concurrently, in addition, this is the same code used by the wildfly CLI tool, so the properties could be even changed "externally" - this is why I prefer adding the synchronized call on each loop.
On the concrete test scope it was failing due to both tests for users and group managements relies on that properties files, and in addition there is some "update" test as well rather than the delete, so at some point that properties files (copied in a temporary folder for the test executions), can be accessed by different tests, and tbh I prefer not to rely on assuming that all tests run in a single thread, as on jenkins they do not...

csadilek:
@romartin I am ok merging this but I am somewhat skeptical about the cause. The tests were failing for you locally as well (before this fix) and there they ran in sequence, right? So the problem also appears in the non-parallel case, without another thread being involved. The original ConcurrentModificationException would confirm that, as there "concurrent" != "parallel".

romartin:
@csadilek good point again - yes the test were failing in a supposed non-parallel execution, but that's what I said before too, as there exist 2 unit tests that points same group files ( users and groups tests ), both unit test  classes do so some stuff on `@BeforeClass`, `@AfterClass` and `@Before`, tbh I'm not sure how junit deals with the calls on both test case instances and intializations/destructions, so I just trying not to make crazy with that kind of eyncrhonization suff, try to fix the syncronization issue, that in fact was a bug, the code was not right, and this finally seems to fix that issue on test scope as well.
So yeah I'm agree with you that still not 100% sure what the cause on my local environment of that failure :( the fix is anyway good, but honestrly I'm still missing somwthing, that's true... WDYT?

PR #631, opened 2017-02-14
proposal that just removes Thread.sleep - but this time keeps the Observer

porcelli:


porcelli:
@manstis @csadilek @ederign just removing Thread.sleep

PR #634, opened 2017-02-23
Ensure jquery is injected.

manstis:
```UberfireSimpleTable``` uses jQuery to set CSS; however there is no guarantee jQuery has actually been loaded into the host page by whatever webapp is using this Widget.


manstis:
@ederign Would you mind reviewing please? Stunner's _showcase_ webapp does not run as ```UberfireSimpleTable``` tries to call jQuery to set some CSS. Other webapps are probably running OK as they're using some other dependencies that ensure jQuery is injected (the "Plugin extensions" for example).

ederign:
+1 thanks for this @manstis

PR #635, opened 2017-02-24
Remove duplicated freemarker AbstractGenerator

psiroky:
Those two IDEA issues are no longer relevant as we don't use the
maven-shade-plugin anymore. We now properly depend on freemarker.


psiroky:
@ederign or @paulovmr could you please take a look?

PR #636, opened 2017-03-02
Upgrade IP BOM from 7.0.0.CR7 to 7.0.0.CR8

psiroky:


uberfire-server/src/test/java/org/uberfire/server/locale/GWTLocaleHeaderFilterTest.java
```diff
@@ -101,7 +101,7 @@ public void testNonExistentLocaleParameter() throws IOException, ServletExceptio
         final HttpServletResponse resp = mock( HttpServletResponse.class );
         final FilterChain chain = mock( FilterChain.class );
 
-        when( req.getParameter( "locale" ) ).thenReturn( "xxx" );
+        when( req.getParameter( "locale" ) ).thenReturn( "xxx_xxx" );
```
ederign: Why did u have to make this change?

psiroky: See the commit message: CR8 upgrades commons-lang3 from 3.1 to 3.4. The change in the test
from "xxx" to "xxx_xxx" is to "fool" the validation of lang3's LocaleUtils.

PR #637, opened 2017-03-02
Fix unit tests when run on MS-Windows

bbrodt:
See https://issues.jboss.org/browse/UF-463

Replaces PR https://github.com/uberfire/uberfire/pull/625

### DIFF
--- a/uberfire-nio2-backport/uberfire-nio2-model/src/main/java/org/uberfire/java/nio/base/dotfiles/DotFileUtils.java
+++ b/uberfire-nio2-backport/uberfire-nio2-model/src/main/java/org/uberfire/java/nio/base/dotfiles/DotFileUtils.java
@@ -16,6 +16,8 @@
 
 package org.uberfire.java.nio.base.dotfiles;
 
+import java.io.File;
+import java.io.FileOutputStream;
 import java.io.OutputStream;
 import java.io.Serializable;
 import java.util.HashMap;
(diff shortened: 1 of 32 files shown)


bbrodt:
@ederign any progress with this PR?

ederign:
@bbrodt first of all thanks for your PR, but unfortunately, I'm not comforted to give a +1 on this. I'll ask for a second opinion from @porcelli and @manstis.

The reason for my argument is that,  at least in my opinion, your PR includes a lot of conditional statements and also a lot of blur logic to a sensible area - uberfire filesystem just to fix tests. That could lead us maintenance issues in future.

Some examples are AbstractPath.getSeparator(), AbstractPath.clearCache, GeneralPathImpl.defaultDirectory and three versions of DotFileUtils.buildDotFile.

Also, why don't use the @porcelli's solution for this? https://gist.github.com/anonymous/fcb57c09cd1805d86ef5a9e1f467d365

uberfire-nio2-backport/uberfire-nio2-model/src/main/java/org/uberfire/java/nio/base/dotfiles/DotFileUtils.java
```diff
@@ -36,11 +38,40 @@ private DotFileUtils() {
...
 
-    public static boolean buildDotFile(final Path path,
-                                       final OutputStream out,
-                                       final FileAttribute<?>... attrs) {
+    public static boolean buildDotFile( final Path path, final FileAttribute<?>... attrs ) {
+        Path dotPath = dot( path );
+        OutputStream out = path.getFileSystem().provider().newOutputStream( dotPath );
         boolean hasContent = false;
-        if (attrs != null && attrs.length > 0) {
+        try {
```
manstis: @bbrodt Given all the hard effort you're taken to ensure streams are closed; you might want to ensure the stream created here (and in related methods) is closed in a finally block (or use try-with-resources).

PR #638, opened 2017-03-06
Revert "Upgrade IP BOM from 7.0.0.CR7 to 7.0.0.CR8 (#636)"

mbiarnes:
This reverts commit 9019eea62d12cd28d03015a5eaa982907ce3bd70.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -22,7 +22,7 @@
     <groupId>org.jboss.integration-platform</groupId>
     <artifactId>jboss-integration-platform-bom</artifactId>
     <!-- Keep in sync with the parent version of uberfire-bom/pom.xml -->
-    <version>7.0.0.CR8</version>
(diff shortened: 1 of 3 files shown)


mbiarnes:
@psiroky for 7.0.0.Beta7 release in kie-group and dashbuilder it is used jboss-ip-bom 7.0.0.CR7 - for the 7.0.0.Beta7 release we should downgrade the jboss-ip-bom used in uberfire to the same version of ip-bom in other projects = 7.0.0.CR7

psiroky:
I don't think we need to revert this, unless there is some specific issue I am not aware about.

In case the revert is indeed needed, lets do it on the release branch. No need to do it here as the next step we would upgrade again to CR8 anyway.

mbiarnes:
@psiroky OK - I will revert this on the release branch of uberfire r1.0.0.Beta7

PR #642, opened 2017-03-09
Ensure UiBinder is instantiated before the instance of BusyPopup.

manstis:
PR https://github.com/droolsjbpm/kie-wb-common/pull/711 has a failing test caused by the instantiation order of the ```BusyPopup``` "View" and ```BusyPopup``` "instance". The "View" needs to be instantiated before the popup "instance".

### DIFF
--- a/uberfire-extensions/uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/common/BusyPopup.java
+++ b/uberfire-extensions/uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/common/BusyPopup.java
@@ -31,8 +31,8 @@ import org.uberfire.ext.widgets.common.client.animations.LinearFadeOutAnimation;
  */
 public class BusyPopup extends DecoratedPopupPanel {
 
-    private static final BusyPopup INSTANCE = new BusyPopup();
     private static LoadingViewBinder uiBinder = GWT.create(LoadingViewBinder.class);
+    private static final BusyPopup INSTANCE = new BusyPopup();
 
     private static Timer deferredShowTimer = new Timer() {
         @Override


manstis:
@ederign @porcelli Please review at your convenience.

uberfire-extensions/uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/common/BusyPopup.java
```diff
@@ -31,8 +31,8 @@
  */
 public class BusyPopup extends DecoratedPopupPanel {
 
-    private static final BusyPopup INSTANCE = new BusyPopup();
     private static LoadingViewBinder uiBinder = GWT.create(LoadingViewBinder.class);
+    private static final BusyPopup INSTANCE = new BusyPopup();
```
ederign: could you please change new BusyPopup() to GWT.Create?

ederign: could you please change new BusyPopup() to GWT.Create?

manstis:
@ederign I can.. but it really opens a can of worms (it's what I've been doing since your change request) 

Changing to use ```GWT.create(BusyPopup.class)``` means GWT Mockito Unit Tests now use a mock ```BusyPopup``` for ```INSTANCE```. Consequentially private member properties (w/o "getters" and "setters") are ```null``` on the mock which leads to NPEs on tests in other modules that touch on ```BusyPopup```.

I _can_ refactor ```BusyPopup`` to be better used in tests; but at the moment I simply swapped two lines around. I didn't see a full-on refactor of ```BusyPopup``` within the scope of trying to get the tests in another PR to pass. Is it OK if I raise a UF JIRA to refactor ```BusyPopup``` instead?

manstis:
I've changed the failing test to work differently; not needing this change.

ederign:
are u sure? in that case I can merge this.

psiroky:
The nightly build shows 14 test failures (in different repos) which awfully look like related to this.

manstis:
@ederign Yes I was sure; I made the requested change and then hit a ton of problems.

@psiroky OK, I'll re-open this and let @ederign decide if he wants to merge "as is".

It's weird why this started causing issues today (AFAIK Eder only reformatted code yesterday, which you'd like to have thought would not shuffle member order.. or perhaps it did!! Uh oh!).

psiroky:
TBH relying on order of initialization of static class members does not seems like a good idea :)

ederign:
If open a can of worms to use gwt.create, let's keep the can closed! ;)

My formatting was just on codestyle + optimize imports. (but was automated by IDEA).

manstis:
@ederign @psiroky Yikes.. it looks like code [formatting](https://github.com/uberfire/uberfire/commit/f8f3737a854dc64f1075b076514c8630bf3da5f2#diff-a3f1bcef9f11cca5e8b5d524db0f903fR34) did re-order things... hopefully this is the last (only) issue!

PR #651, opened 2017-03-30
RHBPMS-4625 :Task Filter List accepts HTML in the Name field which is…

nmirasch:
… rendered when deleting it
Added the SafeHtml conversion to the tab delete confirmation popup and to the filter title to avoid the html interpretation


ederign:
ok to test

ederign:
@nmirasch thanks for your PR. Accepting HTML is this fields is a requirement right?

nmirasch:
@ederign the only reported issue is that html shouldn't be interpreted to avoid entry points for XSS attacks.

PR #668, opened 2017-04-20
JBPM-5767 and JBPM-5894 - Ensure preference loader is started before …

cristianonicolai:
…any dependent bean using @Startup


cristianonicolai:
@paulovmr also related to https://github.com/kiegroup/jbpm-wb/pull/655

PR #671, opened 2017-04-24
JBPM-5542: Stunner - Properties Panel and Process Explorer are shown for incorrect asset

pefernan:
The idea is refreshing only the affected docks when new dock elements are added or removed.

@ederign can you take a look? Does it fit what we talked yesterday?


pefernan:
jenkins retest this

PR #676, opened 2017-04-27
UF-511: Ignore JGitFileSystemProviderSSHTest on IBM JDK

yurloc:
I believe the test only fails because of wrong setup. I tried to add `org.bouncycastle:bcprov-jdk15on:1.56` as a test dependency according to https://bugzilla.redhat.com/show_bug.cgi?id=1192072 but I wasn't successful so I suggest to ignore the test on IBM JDK until the proper setup is implemented. This will make QE life easier because we run the test on different JDKs and there are currently multiple failures related to platform-dependent tests.

Please cherry-pick to 1.0.x and master, too.


yurloc:
1.0.x #678
master #679

PR #681, opened 2017-05-04
[0.9.x] RHBPMS-4706: [GSS](6.4.z) Embedded Authoring perspective is broken when using parameter controlled URL

manstis:
See https://issues.jboss.org/browse/RHBPMS-4706

This PR moves instantiation of the "Menu bar" to a helper to provide a different implementation for "default" vs "standalone" modes of operation (i.e. when workbench URL contains the ```standalone=true``` parameter). Tried and tested with (6.5.x) ```kie-wb``` and ```kie-drools-wb``` both "default" and "standalone" modes.. all OK.

### DIFF
--- a/uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/workbench/widgets/menu/WorkbenchMenuBarPresenter.java
+++ b/uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/workbench/widgets/menu/WorkbenchMenuBarPresenter.java
@@ -19,10 +19,6 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
-import javax.annotation.PostConstruct;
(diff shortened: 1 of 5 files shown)


cristianonicolai:
Seems like the right approach to me, splitting up the presenters based on the mode. We might have to split up it even further if we think that views would have different UI elements but that could be a further enhancement.

manstis:
@cristianonicolai @ederign Could you please review and merge?

manstis:
ok to test

ederign:
@manstis I think the same issue happens on master. Are you planning to also do this there?

manstis:
@ederign Yes.. just getting 0.9.x done first (as there's a customer shouting for it on Product 6.4.x)

ederign:
ok to test

ederign:
testing locally.

ederign:
Green on my machine.

manstis:
Thanks @ederign

PR #688, opened 2017-05-08
Temporarily ignoring the intermittent test 

ederign:
https://issues.jboss.org/browse/UF-512

@manstis @porcelli @paulovmr 

This test is randomly failing on Jenkins and avoiding us to having green builds.

I don't like to do that, but as I know that everyone is busy at the moment, I would like to temporarily ignore this test.


ederign:
See how the build is green now? 😢

manstis:
@ederign @paulovmr @porcelli Theses tests were re-written using Byteman to remove a ```Thread.sleep(..)``` for an arbitrary delay as such delay was not guaranteed to be sufficient in some environments (whilst massively long in others). So, if the test is now failing IMO we should look why as it should no longer fail sporadically; but only for a good reason.

ederign:
Yes. I hope someone has time to take a look on this soon. But for now, I've created the JIRA for this issue and merge this. Red builds really disturbs me.

ederign:
Closing this after talking with @porcelli because can be a real issue.

@manstis I'll try to prioritize this with kris and edson.

PR #690, opened 2017-05-09
Move FreeMarker Configuration to static

manstis:
Move FreeMarker ```Configuration``` to ```static``` to avoid multiple instance creation by concrete ```AbstractGenerator``` subclasses. Include a ```MultiTemplateLoader``` to accept incremental declaration of additional ```TemplateLoader``` for each successive annotation processor within Uberfire using the singleton ```Configuration``` instance.

FreeMarker's documentation states ```Configuration``` should be a singleton for the whole application and shared across threads (objects derived from ```Configuration``` are thread-safe). There remain sporadic compilation [failures](https://kie-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/kie-all-master/job/kie-all-build-master/721/console) caused by what appears to be multiple attempts to access configuration files from multiple invocation (instantiation?) of ```Configuration``` static initializer block.

I've also ensured we construct an instance of ```Construction``` for the most recent version of FreeMarker (and hence, based on their JavaDoc, should contain the most bug fixes). The old approach instantiated an instance compatible with 2.3.0 (not 2.3.25).

Local compilation works fine.. only one real way to see if this fixes any of our parallel build problems.....

### DIFF
--- a/uberfire-workbench/uberfire-workbench-processors/src/main/java/org/uberfire/annotations/processors/AbstractGenerator.java
+++ b/uberfire-workbench/uberfire-workbench-processors/src/main/java/org/uberfire/annotations/processors/AbstractGenerator.java
@@ -19,8 +19,10 @@ import javax.annotation.processing.ProcessingEnvironment;
 import javax.lang.model.element.Element;
 import javax.lang.model.element.PackageElement;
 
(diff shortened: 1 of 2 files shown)


manstis:
@psiroky @mbiarnes @ederign Thoughts and opinions welcome :-)

PR #694, opened 2017-05-13
Avoid unnecessary repetitive calls to Pattern.compile()

jhrcek:
@porcelli please take a look and consider merging this.

Code using `<some string>.matches("regex")` internally calls `Pattern.compile(regex).matcher(input).matches()` The regex compilation step is much more expensive compared to matching step. Caching the Pattern (or avoiding regexes in some cases) will improve performance in these often-called cases.

### DIFF
--- a/uberfire-extensions/uberfire-apps/uberfire-apps-client/src/main/java/org/uberfire/ext/apps/client/home/components/popup/DirectoryNameValidator.java
+++ b/uberfire-extensions/uberfire-apps/uberfire-apps-client/src/main/java/org/uberfire/ext/apps/client/home/components/popup/DirectoryNameValidator.java
@@ -19,9 +19,12 @@ package org.uberfire.ext.apps.client.home.components.popup;
 import org.uberfire.ext.apps.api.Directory;
 import org.uberfire.ext.apps.client.resources.i18n.CommonConstants;
 
+import java.util.regex.Pattern;
+
 public class DirectoryNameValidator {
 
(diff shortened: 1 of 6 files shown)


jhrcek:
ok to test

jsoltes:
@porcelli it looks like at least kie-wb-common is now failing because of this pull-request. The problem is that java.util.regex.Pattern is not gwt-compilable. It's used in NameValidator class in stunner-showcase (see https://kie-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/kie-wb-common-pullrequests/1180/console). Could you please revert this commit? And also could we do something to prevent people doing this change here (the build should not pass).

porcelli:
@jsoltes done!

@psiroky would be possible we change our current jenkins job.. to always build the full stack? so we can avoid such issues

jsoltes:
@porcelli thanks!

psiroky:
@porcelli I am not sure that would be a good thing. Running the full stack takes more than 2.5 hours and usually at least one of the tests fails. That being said, I am planning to add an on-demand "full downstream build" job, which would do this when trigger by a specific phrase.

IMO issues like this should be detected directly by UberFire build. All the -client jars should be used in the showcase, so the compilation should fail directly there.

porcelli:
@psiroky I see your point, but without it... with current codebase... the feedback we have isn't the most useful thing. Maybe we need more machinery... or we just wait longer.

psiroky:
@porcelli I definitely agree something needs to be done. There needs to be reasonably fast feedback loop. I am unfortunately not sure how to easily achieve that.

PR #700, opened 2017-05-16
Avoid unnecessary Pattern.compile() in AbstractPath

jhrcek:
Sorry for the trouble this commit caused in https://github.com/AppFormer/uberfire/pull/694. I didn't realize that client-side doesn't support that. That being said I still think that this server-side change in AbstractPath is worth changing as I figure thousands of Path instances are created during normal operation of kie-wb. @porcelli WDYT?


ederign:
jenkins retest this

ederign:
@manstis now this build is green! ;)

PR #684, opened 2017-05-05
RHBPMS-4707: Embedded Authoring perspective is broken when using parameter controlled URL

manstis:
See https://issues.jboss.org/browse/RHBPMS-4707

This is the corollary of https://github.com/AppFormer/uberfire/pull/681 for master.

(cherry picked from commit ea4477afce4313294c1dc3c926322e4717f0c376) - RHBPMS-4706

PR #705, opened 2017-05-17
[1.0.x] RHBPMS-4707: Embedded Authoring perspective is broken when using parameter controlled URL

manstis:
See https://issues.jboss.org/browse/RHBPMS-4707

This is the corollary of https://github.com/AppFormer/uberfire/pull/684 for 1.0.x (7.0.0.Final)


manstis:
@paulovmr @ederign @krisv @Rikkola Backported as per emails.

Rikkola:
@manstis Thank you for this fix.

PR #706, opened 2017-05-18
Move common PatternFly related Errai UI components from jbpm-wb to ub…

cristianonicolai:
…erfire


ederign:
@cristianonicolai thanks for this. I think that will be the beginning of our PF/BS Library ;)

Could you please add tests for this PR?

cristianonicolai:
@ederign that's the idea :)

tests added.

PR #712, opened 2017-05-19
JBPM-5990 - PagedTable: allow customization of top level toolbar

cristianonicolai:


uberfire-extensions/uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/tables/PagedTable.ui.xml
```diff
@@ -47,6 +47,9 @@
   </ui:style>
 
   <g:FlowPanel addStyleNames="{style.dataGridContainer}">
+    <b:Row>
```
manstis: Could you confirm this has zero height by default? Those **not** using a custom top toolbar probably don't want white space.

cristianonicolai: @manstis that's correct, adding a row and col-md-12 divs won't create any extra height.

PR #713, opened 2017-05-22
RHBPMS-4609: Not able to clone BPMS server git repository over ssh protocol over WAS server.

markcoble:
NOTE:  this is an early PR to get feedback on design/approach taken.  GroupAdapterAuthorizationSource has not been refactored so no need to review with the initial PR.

Currently the getRoles() method of GroupAdapterAuthorizationSource expects an exact match of role name provided in the applications web.xml and the principal name passed to it.  Debugging of the code for this JIRA exposed the fact that certain use cases roles are not being authorized for user/groups where expected.  For instance, with WAS using LDAP a git clone request with ssh does not get the filtering of an LDAP DN entry that occurs for web clients.  So the principal name that is passed to getRoles() is a full DN entry (like cn=admin, ou=groups, dc=example, dc=com).  

The new code committed will allow for mapping application roles to user/groups from any number of realms.  It can also use a simple regex expression for cases like that raised my this JIRA.  That is, a regex match on cn={role} would work.  Regardless of the mapping applied,  a system property can be set specifying the location of mapping providers.

GroupAdapterAuthorizationSource


ederign:
ok to test

uberfire-backend/uberfire-backend-server/src/test/resources/role-mapping-providers.properties
```diff
@@ -0,0 +1,3 @@
+org.uberfire.ldap.role_mapping=config/ldap-role-mapping.json
```
porcelli: -1 for a config file that points to different config files

markcoble: I see it as part of this PR.  it shows up under files change in my view.

uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/security/RoleMapper.java
```diff
@@ -0,0 +1,122 @@
...
+        WebAppListener.registerOnStartupCommand(new Command() {
+
+            @Override
+            public void execute() {
+                registerRolePrincipals();
+            }
+        });
+    }
+
+    public void registerRolePrincipals() {
```
markcoble: The simplest solution that would fix some cases is to have a System property that contains a regex expression that would allow complex group names (like the one LDAP on WAS returns) to be filtered and match a role accordingly.  So this is the answer to your question below about why regex is needed.  If we adopt a more complex role-to- principals mapping approach, then rather than simply defining a regex expression as a System property it could be considered another type of 'mapping provider'.   

If we only do a System property with regex, it seems to me this will become restrictive as well for certain scenarios, 
 If both user and groups from various security realms (ldap, file, etc.)  could map to a role and how the user/group is returned as a principal, then having a clear and direct way of mapping a role to whatever entity can have that role seems advantageous.

uberfire-backend/uberfire-backend-server/src/test/resources/role-mapping-providers.properties
markcoble: The idea here was that there could be one-to-many role mapping providers.  I wouldn't want to configure each as a System property.

porcelli: sorry, got it

uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/security/RoleMapper.java
porcelli: I see your point..  but in the other hand... add all this flexibility in a bug fix release worries me. For this specific jira, would the ldap config would be enought to fix the issue?

uberfire-backend/uberfire-backend-server/src/test/resources/role-mapping-providers.properties
porcelli: I hope that we could make this issue fixed with just the ldap config.. and we wouldn't need the config file that points to other config files.

uberfire-backend/uberfire-backend-server/pom.xml
```diff
@@ -130,6 +130,17 @@
       <artifactId>powermock-api-mockito</artifactId>
       <scope>test</scope>
     </dependency>
+    
+    <!-- Marshalling -->
+    <dependency>
+      <groupId>org.json</groupId>
+      <artifactId>json</artifactId>
```
porcelli: this still worries me... hopefully we could just use properties

markcoble:
@porcelli Putting my comments together.  I will get the exact user/groups Citi is using.  I believe a simple system property with a regex pattern defined will solve their issue.  The design I am proposing stemmed from conversations with Maciej (see some of his comments in the issue.)   I also am thinking of our drive to be cloud ready!  I  struggled with a property which is a location pointing to other locations!  However, with Openshift these are just config maps or secrets accessed via a shared mounted volume or similar. It also would allow the mapping data to be accessed and updated via Rest API.  That said, I can do the simple approach since this will be a fix to 6.5.x and hopefully get support for the more complicated mapping provider approach in 7.x!

PR #715, opened 2017-05-22
[0.9.x][RHBPMS-4669] filter out invalid (null) login principals

psiroky:
In some cases (like one the described in the JIRA), the list of
principals contained a principal with 'null' name, which is
probably wrong. This caused a NPE further down as the name was used
to compute hashcode. Filtering the invalid principals seems to be
the easiest way to workaround that issue.

Backport of #707.


porcelli:
assuming that we didn't have any test before, right?

psiroky:
The test is in uf-extensions: https://github.com/AppFormer/uberfire-extensions/pull/320

porcelli:
ops :P - thank you!

PR #717, opened 2017-05-24
UF-517: Boostrap Selector CSS overriding Patternfly Selector CSS

manstis:
See https://issues.jboss.org/browse/UF-517

@cristianonicolai IDK if this is where/how we should override incompatibilities between BS + PF CSS?

### DIFF
--- a/uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/resources/org/uberfire/client/views/static/uberfire-patternfly.css
+++ b/uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/resources/org/uberfire/client/views/static/uberfire-patternfly.css
@@ -338,3 +338,8 @@
 .uf-perspective-context-menu-empty {
     padding-top: 43px;
 }
+
+/*https://issues.jboss.org/browse/UF-517*/
+.bootstrap-select.btn-group .btn .caret {
+    margin-top: -4px !important;
+}


manstis:
@cristianonicolai You're probably best positioned to review?!

cristianonicolai:
@manstis I think the real issue is due to the GWTBootstrap library that includes all  bootstrap-select resources again once it is used. Overriding what we define in Uberfire PatternFly. Your fix here obliviously works fine but another way to fix this would be to ensure that resources from the GWTBootstrap library aren't loaded. That would mean using the `SelectNoResources` GWT module instead of `org.gwtbootstrap3.extras.select.Select`. I'm just not sure how practical this change would be, I believe we would need to ensure that this isn't used anywhere in the workbench.

manstis:
Thanks for the valuable feedback @cristianonicolai I'll investigate usage.

manstis:
OK, see https://github.com/gwtbootstrap3/gwtbootstrap3-extras/issues/336

It seems like the ```SelectNoResources``` module doesn't include the minimal required JavaScript either :-(

What's our choices for this scenario!? I didn't really want to have to go to the lengths of bundling the JS into a new(?UF) module - like [this](https://github.com/AppFormer/uberfire/tree/master/uberfire-workbench/uberfire-workbench-client-views-patternfly) - and we don't really want to include in the different webapps either.. I wonder if 2px offset is really worth the effort ;-)

manstis:
@cristianonicolai I see ```uberfire-workbench-client-views-patternfly``` already packages [bootstrap-select](https://github.com/AppFormer/uberfire/blob/master/uberfire-workbench/uberfire-workbench-client-views-patternfly/pom.xml#L202) but I don't see the resulting JS file "in" any webapp... I'd appreciate any assistance you can give pointing me to how best consume this JS?

manstis:
@cristianonicolai Sorry, Mr Thickie again.. think I'm OK.. just looking at ```PatternFlyBootstrapper.ensureBootstrapSelectIsAvailable()``` :-)

manstis:
OK, eventually I did the correct fix :-)

Thanks for your guidance @cristianonicolai

cristianonicolai:
@manstis you welcome :)

PR #728, opened 2017-06-01
JBPM-5990 - Improvements to Popover element

cristianonicolai:
Fixing issues to load the data-* attributes properly.

### DIFF
--- a/uberfire-showcase/uberfire-webapp/src/main/java/org/uberfire/client/screens/widgets/WidgetsShowcaseView.html
+++ b/uberfire-showcase/uberfire-webapp/src/main/java/org/uberfire/client/screens/widgets/WidgetsShowcaseView.html
@@ -1,0 +1,29 @@
+<div data-field="root">
+    <div class="col-md-12">
+        <section>
+            <h3>Popover examples</h3>
+            <p>See
+                <a href="http://www.patternfly.org/pattern-library/widgets/#popover">http://www.patternfly.org/pattern-library/widgets/#popover</a> for complete popover documentation.
+            </p>
+            <ul class="list-group">
+                <li class="list-group-item">
+                    <h4 class="list-group-item-heading">Example using data-* attributes</h4>
+                    <a class="list-group-item-text btn btn-link" role="button" data-toggle="popover" data-field="popover" data-trigger="hover" data-content="test..."title="" data-placement="left" data-container="body">
+                        <span class="pficon pficon-info"></span>
+                    </a>
+                </li>
+                <li class="list-group-item">
+                    <h4 class="list-group-item-heading">Example using dynamic element</h4>
+                    <div class="list-group-item-text" data-field="dynamic-popover"></div>
+                </li>
+                <li class="list-group-item">
+                    <h4 class="list-group-item-heading">Example overriding data-* attributes</h4>
+                    <a class="list-group-item-text btn btn-link" role="button" data-toggle="popover" data-field="popover-override" data-trigger="hover" data-content="test..." title="" data-placement="top" data-container="body">
+                        <span class="pficon pficon-info"></span>
+                    </a>
+                </li>
+            </ul>
+
+        </section>
+    </div>
+</div>
--- a/uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/widgets/Popover.java
+++ b/uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/widgets/Popover.java
@@ -17,22 +17,50 @@
 package org.uberfire.client.views.pfly.widgets;
 
 import javax.annotation.PostConstruct;
+import javax.annotation.PreDestroy;
 import javax.enterprise.context.Dependent;
 import javax.inject.Inject;
 
+import com.google.gwt.core.client.Scheduler;
+import jsinterop.annotations.JsType;
 import org.jboss.errai.common.client.api.IsElement;
 import org.jboss.errai.common.client.dom.Anchor;
 import org.jboss.errai.common.client.dom.HTMLElement;
+import org.uberfire.client.views.pfly.widgets.JQueryProducer.JQuery;
+import org.uberfire.client.views.pfly.widgets.JQueryProducer.JQueryElement;
 
+/**
+ * Wrapper component for PatternFly's <a href="http://www.patternfly.org/pattern-library/widgets/#popover">Popover</a>
+ */
 @Dependent
 public class Popover implements IsElement {
 
     @Inject
     private Anchor anchor;
 
+    @Inject
+    private JQuery<JQueryPopoverElement> jQuery;
+
     @PostConstruct
     public void init() {
-        init(anchor);
+        Scheduler.get().scheduleDeferred(() -> jQuery.wrap(getElement()).popover());
+    }
+
+    @PreDestroy
+    public void destroy() {
+        jQuery.wrap(getElement()).popover("destroy");
+    }
+
+    public void show() {
+        jQuery.wrap(getElement()).popover("show");
+    }
+
+    public void hide() {
+        jQuery.wrap(getElement()).popover("hide");
+    }
+
+    public void toggle() {
+        jQuery.wrap(getElement()).popover("toggle");
     }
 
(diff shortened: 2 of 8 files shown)


uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/widgets/Popover.java
```diff
@@ -50,9 +105,22 @@ public HTMLElement getElement() {
     }
 
     public native void init(final HTMLElement e) /*-{
-        $wnd.jQuery(e).popover({
-            container: 'body',
```
cristianonicolai: @nmirasch I recall you had to add this for some element on case management. Was that the roles building block? I did check that and seems fine but I wanted to double check with you.

uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/widgets/Popover.java
```diff
@@ -17,13 +17,18 @@
...
+import com.google.gwt.core.client.Scheduler;
 import org.jboss.errai.common.client.api.IsElement;
 import org.jboss.errai.common.client.dom.Anchor;
 import org.jboss.errai.common.client.dom.HTMLElement;
 
+/**
+ * Wrapper component for PatternFly's <a href="http://www.patternfly.org/pattern-library/widgets/#popover">Popover</a>
+ */
 @Dependent
 public class Popover implements IsElement {
```
cristianonicolai: @rbarriuso FYI, this element should be useful for your refactor in the ProcessInstanceListView

rbarriuso: Thanks! Actually I'm doing something similar, but using Errai templates, as required by JBPM-6024

uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/widgets/Popover.java
```diff
@@ -32,16 +37,66 @@
...
 
     public void setTitle(final String title) {
         anchor.setTitle(title);
     }
 
     public void setContent(final String content) {
-        anchor.setAttribute("data-content",
-                            content);
+        setDataAttribute("content",
+                         content);
```
rbarriuso: Perhaps it would be interesting to add some kind of `content` preprocessing when the "html mode" is set to `true`, because the generated attribute value can contain confusing quote marks. Something like [this](https://github.com/kiegroup/jbpm-wb/pull/741/files#diff-dd7f023e8d1038741ccd2b8c47d5d4afR119) could help

uberfire-showcase/uberfire-webapp/src/main/java/org/uberfire/client/screens/widgets/WidgetsShowcaseView.html
```diff
@@ -0,0 +1,29 @@
...
+            <ul class="list-group">
+                <li class="list-group-item">
+                    <h4 class="list-group-item-heading">Example using data-* attributes</h4>
+                    <a class="list-group-item-text btn btn-link" role="button" data-toggle="popover" data-field="popover" data-trigger="hover" data-content="test..."title="" data-placement="left" data-container="body">
+                        <span class="pficon pficon-info"></span>
+                    </a>
+                </li>
+                <li class="list-group-item">
+                    <h4 class="list-group-item-heading">Example using dynamic element</h4>
+                    <div class="list-group-item-text" data-field="dynamic-popover"></div>
```
rbarriuso: For some reason I don't see this popover properly, see the following screenshot:
popover2

rbarriuso: Seen on Firefox 53 and Chome 58

cristianonicolai: @rbarriuso thanks, I've fixed some conflicting css styles.

uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/widgets/Popover.java
cristianonicolai: That could be useful but I'm not sure if that is something that belongs to the component or more of a util function. GWT already offers some sanitizing functions, see: [DevGuideSecuritySafeHtml](http://www.gwtproject.org/doc/latest/DevGuideSecuritySafeHtml.html). Perhaps the component should accept only SafeHtml instead of plain String.
@ederign @mbarkley any recommendations here? Would even SafeHtml go away on GWT 3?

rbarriuso: @cristianonicolai The sanitize functions from `SafeHtmlUtils`, like `fromSafeConstant()` or `fromTrustedString()`, can be useful here in order to assure there are not security threats in the injected code.
Nevertheless, even after applying them, I think it would be good to process the quote marks (I haven't found any existing utility function for that). Probably embedding HTML code in a tag attribute is a border case and even not the best practice, and it certainly has its limitations (e.g. cannot nest more levels of code or quote marks), but it's required by the popover element.
I guess an option would be creating our own method in a utility class.

mbarkley: @cristianonicolai I'd suggest we make our own utility method for sanitising data. It can use SafeHtml for the moment, but we don't know if that API will stay around so we should minimize its usage sites.

cristianonicolai: thanks @mbarkley @rbarriuso, sounds like creating a utility class would be the best alternative. I'll leave the component as is then, only receiving a String for the content.

ederign:
@mbarkley @rbarriuso could you please take a look on this?

rbarriuso:
Looks good to me

ederign:
@cristianonicolai could you please take a look on the conflicts?

uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/widgets/Popover.java
```diff
@@ -50,9 +105,22 @@ public HTMLElement getElement() {
...
     public native void init(final HTMLElement e) /*-{
-        $wnd.jQuery(e).popover({
-            container: 'body',
-            trigger: 'focus'
-        });
+        $wnd.jQuery(e).popover();
+    }-*/;
+
+    private native void show(final HTMLElement e) /*-{
+        $wnd.jQuery(e).popover('show');
```
mbarkley: We should not be using JSNI (unless absolutely necessary) and use JS interop for these kinds of things instead as JSNI is likely to disappear in GWT 3.

[Here](https://github.com/errai/errai-tutorial/blob/master/src/main/java/org/jboss/errai/demo/client/local/JQueryProducer.java) is an example of a simple (incomplete) wrapper for JQuery that can be injected, and [here](https://github.com/errai/errai-tutorial/blob/master/src/main/java/org/jboss/errai/demo/client/local/AppSetup.java#L51) is an example usage.

porcelli: +1 for @mbarkley

cristianonicolai: @mbarkley thanks for pointing that out. I've updated my code using a jquery wrapper.

PR #729, opened 2017-06-01
UF-528 - Add missing patternfly.js file and add entry point to bootst…

cristianonicolai:
…rap PatternFly


uberfire-showcase/uberfire-webapp/src/main/java/org/uberfire/client/ShowcaseEntryPoint.java
```diff
@@ -152,8 +146,13 @@
     @AfterInitialization
     public void startApp() {
         hideLoadingPopup();
+        GWT.log("PatternFly version: " + getPatternFlyVerision());
```
ederign: small typo on method name

cristianonicolai:
@ederign done, thanks :)

PR #732, opened 2017-06-06
JBPM-5990 - Add MomentJS library

cristianonicolai:
Requirement for new date range picker widget
Matched available language locales with the ones used by kie-wb


cristianonicolai:
@ederign +1, will take a look on that for future work.

PR #733, opened 2017-06-06
Fix test event compilation error

cristianonicolai:
Removed class ResourceCopiedEventMock to use EventSourceMock instead.


cristianonicolai:
@ederign somehow ResourceCopiedEventMock fails to compile in my local IntelliJ, complaining about a missing a `select(Class<U> subtype, Annotation... qualifiers)` method.

PR #734, opened 2017-06-06
DASHBUILDE-222: Custom persp. is offered in nav tree even after it's deleted

dgutierr:
Fix + tests.

@ederign  Can you review & approve please?


ederign:
thanks david!

PR #738, opened 2017-06-08
JBPM-6069: Missing CRUD buttons in asset editors

manstis:
See https://issues.jboss.org/browse/JBPM-6069

### DIFF
--- a/uberfire-extensions/uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/BaseEditor.java
+++ b/uberfire-extensions/uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/BaseEditor.java
@@ -196,27 +196,16 @@ public abstract class BaseEditor {
         concurrentUpdateSessionInfo = null;
     }
 
-    private void buildMenuBar() {
-        if (menuBuilder != null) {
-            menus = menuBuilder.build();
-        }
-    }
-
     protected void showVersions() {
 
     }
 
     /**
-     * If you want to customize the menu override this method.
+     * If you want to customize the menu content override this method.
      */
     protected void makeMenuBar() {
         if (menuItems.contains(SAVE)) {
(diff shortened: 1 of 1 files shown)


manstis:
@jomarko @ederign Could you please review?

manstis:
Part of an ensemble:

- https://github.com/AppFormer/uberfire/pull/738
- https://github.com/kiegroup/kie-wb-common/pull/926
- https://github.com/kiegroup/drools-wb/pull/525
- https://github.com/kiegroup/optaplanner-wb/pull/177
- https://github.com/kiegroup/jbpm-designer/pull/615

uberfire-extensions/uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/BaseEditor.java
```diff
@@ -241,6 +230,17 @@ public void execute() {
         }
     }
 
+    /**
+     * If you want to customize the menu construction override this method. {@link BaseEditor#makeMenuBar()}
+     * should be used to add items to the {@link BasicFileMenuBuilder}. This method then instructs
+     * {@link BasicFileMenuBuilder#build()} to create the {@link Menus}
+     */
+    protected void buildMenuBar() {
+        if (menuBuilder != null) {
```
jomarko: Just question, I see `menuBuilder` is injected via annotation. Is there possibility that it is null? I mean, am I right that the compilation should fail if there is no bean to inject? But take this just as a question, I am not against this preventive approach. It is definitely better to have this `if` as NPE during runtime.

manstis: @jomarko You're right.. it should never be ```null``` otherwise CDI would complain it cannot satisfy an injection point. I just moved the method to make its order consistent with the order in which it's invoked from the ``init(..)``` method (oh, and made it ```protected``` which is really the jist of the whole change).

PR #740, opened 2017-06-08
UF-527, DROOLS-1582: moving generic maven utils to (soon to be) appformer repository

porcelli:
I know that this structure looks a bit odd (using appformer namespace inside an uberfire module), but this is more or less like the `chicken egg` problem: 
 - in order to have AppFormer repository setup (a merger of UF, Dashbuilder, Guvnor and parts of KIE-WB-Common) we need remove kie specifics depdencies.
 - the maven utils here are the last blocker bits for the merge

FAQ:

Q: Why appformer namespace with a UberFire maven module?
A: Trying to make this move incremental, but less impactant possible. Easy to change pom files, than change all namespace of types used

Q: Why not an AppFormer maven module?
A: Would create a odd situation, as we would have different versions of sources inside the same repository (appformer and uberfire); or would create a Maven modules of appformer in 1.1.0-SNAPSHOT that later one will become 0.9.0-SNAPSHOT.

Links (@mariofusco extracting from kie codebase): 
https://github.com/mariofusco/droolsjbpm-knowledge/commit/7b1fbb0fb55142772f2eff4f980ce26f98d5af4a
https://github.com/mariofusco/drools/commit/4e471f7ddfa9d1000b877090a4cce90223c6f5d6

ps: I was trying to preserve history, but after many attempts I gave up, as I always ended `loosing` it due the radical changes of directories (yes.. I tried `git log --folow` to check)


porcelli:
@ederign @mariofusco @mbarkley @manstis enjoy this `mess` :)

mariofusco:
@porcelli I'll start integrating what you did with kie-api/drools right now. I'll ping you when I'll be ready with this.

PR #742, opened 2017-06-12
[0.9.x][RHBRMS-2839] return empty role/group list instead of null

psiroky:
@romartin could you please take a look?


uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/security/adapter/GroupAdapterAuthorizationSource.java
```diff
@@ -157,13 +155,11 @@ public GroupAdapterAuthorizationSource( final Iterator<GroupsAdapter> groupsAdap
...
                 }
 
-                return result;
-
             }
 
         }
 
-        return null;
+        return result;
```
psiroky: @romartin was there any particular reason to return null instead of empty list here?

PR #746, opened 2017-06-14
configure http headers on aether session

mariofusco:


mariofusco:
@porcelli please review this

PR #747, opened 2017-06-14
[AF-544] Allow close a expanded dock via API

ederign:


ederign:
@paulovmr thanks for the review! fixed!

porcelli:
Wondering about the purpose of the "DeSelect" feature... 

What about toggle feature? If this is about close/expand.. why not use setExpand or just close() or hide?

ederign:
@porcelli deselect and expand API will be used by bookmarkable URLs. Even if we use to expand or close without the reference for the dock, we should pass with dock area we are wanna close.
 
Is that ok for you?

uberfire-extensions/uberfire-simple-docks/uberfire-simple-docks-client/src/main/java/org/uberfire/client/docks/UberfireDocksImpl.java
```diff
@@ -114,6 +112,13 @@ public void expand(UberfireDock dock) {
     }
 
     @Override
+    public void deselect(UberfireDock dock) {
+        if (docksBars.isReady()) {
+            docksBars.deselect(dock);
```
porcelli: what if... is not ready? aren't we loosing the call?

porcelli: what about add to a registry of after initialization/ready?

PR #762, opened 2017-07-05
JBPM-6159 - Fix adjust column width test failure when converting Stri…

cristianonicolai:
…ng with decimal


cristianonicolai:
@nmirasch fixing decimal String problem in here as it started affecting more tests.

PR #764, opened 2017-07-07
AF-547: VFS commit batch broken

mcimbora:
Perform dot file removal & file changes in a single commit by enforcing batch FS state. This simplifies version history in cases when a file is not saved within a batch boundary.

I also noticed maven-surefire-plugin configuration defined a single test only (with the rest skipped). If this is intentional, please let me know.

Part of:
https://github.com/kiegroup/guvnor/pull/473
https://github.com/AppFormer/uberfire/pull/764

### DIFF
--- a/uberfire-nio2-backport/uberfire-nio2-impls/uberfire-nio2-jgit/pom.xml
+++ b/uberfire-nio2-backport/uberfire-nio2-impls/uberfire-nio2-jgit/pom.xml
@@ -131,29 +131,6 @@
           </execution>
         </executions>
       </plugin>
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-surefire-plugin</artifactId>
-        <configuration>
-          <includes>
-            <include>org/uberfire/java/nio/fs/jgit/JGitFileSystemProviderBytemanTest</include>
-          </includes>
-          <useSystemClassLoader>true</useSystemClassLoader>
-          <useManifestOnlyJar>true</useManifestOnlyJar>
-          <forkMode>once</forkMode>
-          <!--<parallel>false</parallel>-->
-          <!-- ensure we don't inherit a byteman jar from any env settings -->
-          <environmentVariables>
-            <BYTEMAN_HOME></BYTEMAN_HOME>
-          </environmentVariables>
-          <systemProperties>
-            <property>
-              <name>org.jboss.byteman.home</name>
-              <value></value>
-            </property>
-          </systemProperties>
-        </configuration>
-      </plugin>
     </plugins>
   </build>
 </project>
--- a/uberfire-nio2-backport/uberfire-nio2-impls/uberfire-nio2-jgit/src/main/java/org/uberfire/java/nio/fs/jgit/JGitFileSystemProvider.java
+++ b/uberfire-nio2-backport/uberfire-nio2-impls/uberfire-nio2-jgit/src/main/java/org/uberfire/java/nio/fs/jgit/JGitFileSystemProvider.java
@@ -1144,37 +1144,59 @@ public class JGitFileSystemProvider implements SecuredFileSystemProvider,
 
         return new SeekableByteChannelFileBasedImpl(new RandomAccessFile(file,
                                                                          "rw").getChannel()) {
+            /**by
+             * The method makes sure a 'dot' file removal and actual file changes are done in a single commit.
+             *
+             * If a batch is already enabled on the filesystem being used, no action is required
+             * If a batch is not enabled on the filesystem, enable it
+             */
             @Override
             public void close() throws java.io.IOException {
                 super.close();
 
                 File tempDot = null;
                 final boolean hasDotContent;
-                if (options != null && options.contains(new DotFileOption())) {
-                    deleteIfExists(dot(path),
-                                   extractCommentedOption(options));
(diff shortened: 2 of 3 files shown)


mcimbora:
@ederign Could you please review the PRs? Thanks.

mcimbora:
Some assets worked just fine, as they are saved withing a batch (e.g. DataObject). Saving other types (e.g. .gdrl Globals definition) resulted in multiple commits.

uberfire-nio2-backport/uberfire-nio2-impls/uberfire-nio2-jgit/src/main/java/org/uberfire/java/nio/fs/jgit/JGitFileSystemProvider.java
```diff
@@ -1231,37 +1231,59 @@ private SeekableByteChannel createANewByteChannel(final Path path,
...
-                    tempDot = File.createTempFile("meta",
-                                                  "dot");
-                    hasDotContent = buildDotFile(path,
-                                                 new FileOutputStream(tempDot),
-                                                 attrs);
-                } else {
-                    hasDotContent = false;
+                boolean batchSet = false;
+                if (!gPath.getFileSystem().isOnBatch()) {
+                    setAttribute(gPath,
```
porcelli: this is dangerous =/
this change an important aspect, and might create a different result. back in time we did this for a purpose, just can't remember what it was.

uberfire-nio2-backport/uberfire-nio2-impls/uberfire-nio2-jgit/src/main/java/org/uberfire/java/nio/fs/jgit/JGitFileSystemProvider.java
```diff
@@ -1231,37 +1231,59 @@ private SeekableByteChannel createANewByteChannel(final Path path,
...
+                               put(gPath.getPath(),
+                                   file);
+                               if (hasDotContent) {
+                                   put(toPathImpl(dot(gPath)).getPath(),
+                                       dotfile);
+                               }
+                           }}));
+                } finally {
+                    if (batchSet) {
+                        setAttribute(gPath,
```
porcelli: if we would go to this route.. what i'm not sure if we should... this should be somehow set to be syncronized and we know that this can't be executed concurrently

uberfire-nio2-backport/uberfire-nio2-impls/uberfire-nio2-jgit/pom.xml
```diff
@@ -137,29 +137,6 @@
           </execution>
         </executions>
       </plugin>
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-surefire-plugin</artifactId>
-        <configuration>
-          <includes>
-            <include>org/uberfire/java/nio/fs/jgit/JGitFileSystemProviderBytemanTest</include>
```
porcelli: this is weird. we should remove it. but we should keep the Byteman configs.. i think

uberfire-nio2-backport/uberfire-nio2-impls/uberfire-nio2-jgit/src/main/java/org/uberfire/java/nio/fs/jgit/JGitFileSystemProvider.java
mcimbora: @porcelli @ederign Copying Q. from guvnordev: Do you see any other feasible way to do this apart from setting batch state on the FS? I experimented with passing new option (e.g. BatchCommitOption) to [commit()](https://github.com/mcimbora/uberfire/blob/494742436769f786fd128807d04cd74c5bd4129b/uberfire-nio2-backport/uberfire-nio2-impls/uberfire-nio2-jgit/src/main/java/org/uberfire/java/nio/fs/jgit/JGitFileSystemProvider.java#L2631) to indicate the commits (file change + dot file removal) should be squashed. Seemed to work fine, however not aware of all the caveats.

porcelli: @mcimbora i understand that this might create a fix for the issue created by recent changes on some parts of VFS.
But I'd rather first fix the issue in the origin.. and then get back here to refine behavior.
The code here would need to be better syncronized to make sure we avoid any racing condition.

mcimbora: @porcelli @ederign Sorry for the delay folks, was busy with other priorities. I updated guvnor-side of the issue https://github.com/kiegroup/guvnor/pull/473, which prevents creation of dot files for non-generated files in the workbench. This alone should resolve JBPM-6078. Can we get the PR merged soon?

As to the generic solution on UF side (this PR), unfortunately I can't provide better solution than this with my current knowledge of jgit internals. Let's make a final decision on this, if the current fix is insufficient, I'll have to hand the issue to you.

Thanks.

mcimbora:
Jenkins retest this.

PR #771, opened 2017-07-12
GUVNOR-3070: Guided Decision Tables: Column Wizard - Edit mode

karreiro:
See https://github.com/kiegroup/drools-wb/pull/541


karreiro:
@paulovmr Could you please take a look?

karreiro:
This PR does not have any dependency with `drools-wb` PR. It can be merged right now. ;-)

PR #788, opened 2017-07-31
[JBPM-6278] NullPointerException when configuring user and group management services using KC adapter provider

lhwong:
Fixed #JBPM-6278 NullPointerException when configuring user and group management services using KC adapter provider


ederign:
Hi @lhwong thanks for this! Could you please add tests for this PR? Also, if this fix JBPM-6278 could you please change your commit message for "[JBPM-6278] NullPointerException when configuring user and group management services using KC adapter provider"

PR #789, opened 2017-08-01
[do not merge] JBPM-6148 - New CDI Profile API to include/exclude beans

cristianonicolai:
New API that switches CDI beans on/off based on the active profile (org.uberfire.profile.active system property).
 If no profile is defined at all, all beans still active.
 Once a profile is active, only the beans for the current profile remain available for CDI.
 Client side code for screens, perspectives, editors, etc is also filtered using Errai's `@EnabledByProperty`

### DIFF
--- a/uberfire-backend/uberfire-backend-api/src/main/java/org/uberfire/backend/profile/ProfileManager.java
+++ b/uberfire-backend/uberfire-backend-api/src/main/java/org/uberfire/backend/profile/ProfileManager.java
@@ -1,0 +1,25 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.uberfire.backend.profile;
+
+import org.jboss.errai.bus.server.annotations.Remote;
+
+@Remote
+public interface ProfileManager {
+
+    boolean isProfileEnabled(String profile);
+}
--- a/uberfire-backend/uberfire-backend-cdi/src/main/java/org/uberfire/backend/server/cdi/profile/ProfileManagerImpl.java
+++ b/uberfire-backend/uberfire-backend-cdi/src/main/java/org/uberfire/backend/server/cdi/profile/ProfileManagerImpl.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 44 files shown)


cristianonicolai:
Depends on https://github.com/errai/errai/pull/276

uberfire-backend/uberfire-backend-api/src/main/java/org/uberfire/backend/profile/ProfileManager.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package org.uberfire.backend.profile;
+
+import org.jboss.errai.bus.server.annotations.Remote;
+
+@Remote
+public interface ProfileManager {
+
+    Boolean isProfileEnabled(String profile);
```
manstis: Did you want tri-state? ```true```, ```false``` or ```null```?

uberfire-backend/uberfire-backend-cdi/src/main/java/org/uberfire/backend/server/cdi/profile/ProfileManagerImpl.java
```diff
@@ -0,0 +1,48 @@
...
+@ApplicationScoped
+public class ProfileManagerImpl implements ProfileManager {
+
+    protected static final String PROFILE_SYSTEM_PROPERTY_NAME = "org.uberfire.profile.active";
+
+    public static Boolean isProfileActive(final String profile) {
+        if (isNullOrEmpty(profile)) {
+            return false;
+        }
+        final String activeProfiles = System.getProperty(PROFILE_SYSTEM_PROPERTY_NAME);
```
manstis: ```activeProfiles``` implies you can have more than one profile active; yet your check suggests this should be ```activeProfile```

uberfire-backend/uberfire-backend-cdi/src/test/java/org/uberfire/backend/server/cdi/profile/ProfileManagerImplTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+    @InjectMocks
+    ProfileManagerImpl profileManager;
+
+    @Test
+    public void testProfileActive(){
+        assertFalse(profileManager.isProfileEnabled(null));
+        assertFalse(profileManager.isProfileEnabled(""));
+        assertTrue(profileManager.isProfileEnabled("runtime"));
+        System.setProperty(PROFILE_SYSTEM_PROPERTY_NAME, "authoring");
```
manstis: You might want to roll back setting the System Property after the test in case other tests (now or in the future) depend on the setting (and probably depend on it **not** being set).

uberfire-workbench/uberfire-workbench-processors-tests/src/test/java/org/uberfire/annotations/processors/PerspectiveProcessorTest.java
```diff
@@ -465,6 +465,42 @@ public void testQualifiersInTheActivity() throws FileNotFoundException {
                      result.getActualCode());
     }
 
+    @Test
```
manstis: How about a test for "not in the profile" too?

cristianonicolai: @manstis not sure exactly what you mean, in this case, I'm only checking the code generation part and not if there is a profile active.

uberfire-backend/uberfire-backend-api/src/main/java/org/uberfire/backend/profile/ProfileManager.java
cristianonicolai: Changed to `boolean`.

uberfire-backend/uberfire-backend-cdi/src/main/java/org/uberfire/backend/server/cdi/profile/ProfileManagerImpl.java
cristianonicolai: Fixed

uberfire-backend/uberfire-backend-cdi/src/test/java/org/uberfire/backend/server/cdi/profile/ProfileManagerImplTest.java
cristianonicolai: Fixed

PR #811, opened 2017-08-22
Remove 'bundle' packaging where not needed

psiroky:
The 'bundle' packaging is only needed when using the artifacts in OSGi
containers. None of the touched modules is being used in that way.


porcelli:
Jenkins please execute full downstream build

PR #813, opened 2017-08-24
Upgrade freemarker to 2.3.26.jbossorg-1

psiroky:
As far as I have been able to test, this fixes the random compilation errors with annotation processors (AF-600). The disadvantage is that we are now using "custom" version (even if with just trivial changes). I am planning to propose the same workaround for the freemarker upstream, but at this point I am not sure it would be accepted since it is really is just a workaround.

This custom version is present in the JBoss.org repo and as such should not be pushed (and won't be) pushed into Maven Central. However, that should not really be a problem as we hardcode the jboss.org repo into our poms anyway.

@porcelli, @ederign please take a look and let me know what you think.


porcelli:
@psiroky go ahead and merge! of course we should try to get this into freemarker as much as possible!

PR #816, opened 2017-08-29
Build fixes

psiroky:


psiroky:
The plugin upgrade unfortunately does not solve the occasional NPE :(

PR #821, opened 2017-09-05
GUVNOR-3344: [Guided Decision Table] Editing BRL columns does not work

karreiro:
See https://issues.jboss.org/browse/GUVNOR-3344

Before:
before

After:
after

---

Part of an ensemble:
- https://github.com/AppFormer/uberfire/pull/821
- https://github.com/kiegroup/drools-wb/pull/595


karreiro:
Hi @manstis and @jomarko,

I'm closing this PR and opening another one.

Instead of "adjusting" the indexes that were being messy by the update process. I found what was messing them.
(I needed to do this because the reindex approach had some undesirable collateral effects).

During the update process of a column, we're creating new columns *and* deleting the old ones. The problem is that the wrong columns were being removed.

Now everything should be fixed by this PR (there's a reproducer there too):
https://github.com/AppFormer/uberfire/pull/845

:-)

PR #826, opened 2017-09-12
GUVNOR-3468: GridLienzoPanel/GridLienzoScrollUI: Improve re-usability

karreiro:
See https://issues.jboss.org/browse/GUVNOR-3468

guvnor-3468

---

Part of an ensemble:
- https://github.com/AppFormer/uberfire/pull/826
- https://github.com/kiegroup/kie-wb-common/pull/1158
- https://github.com/kiegroup/drools-wb/pull/611

### DIFF
--- a/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/layer/impl/GridLienzoPanel.java
+++ b/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/layer/impl/GridLienzoPanel.java
@@ -44,7 +44,9 @@ public class GridLienzoPanel extends FocusPanel implements RequiresResize,
 
     private final AbsolutePanel internalScrollPanel = new AbsolutePanel();
 
-    private final AbsolutePanel mainPanel = new AbsolutePanel();
+    private final AbsolutePanel scrollPanel = new AbsolutePanel();
+
+    private final AbsolutePanel rootPanel = new AbsolutePanel();
 
     private final GridLienzoScrollHandler gridLienzoScrollHandler;
 
@@ -75,26 +77,38 @@ public class GridLienzoPanel extends FocusPanel implements RequiresResize,
             }
         });
 
-        domElementContainer.setPixelSize(width,
-                                         height);
+        updatePanelSize(width,
+                        height);
     }
 
     private GridLienzoPanel(final LienzoPanel lienzoPanel) {
         this.lienzoPanel = lienzoPanel;
         this.gridLienzoScrollHandler = new GridLienzoScrollHandler(this);
 
-        setupMainPanel();
+        setupPanels();
         setupScrollHandlers();
         setupDefaultHandlers();
     }
 
-    void setupMainPanel() {
+    void setupPanels() {
+        setupScrollPanel();
+        setupDomElementContainer();
+        setupRootPanel();
+
+        add(getRootPanel());
+    }
+
+    void setupScrollPanel() {
+        getScrollPanel().add(getInternalScrollPanel());
+    }
+
+    void setupDomElementContainer() {
         getDomElementContainer().add(getLienzoPanel());
+    }
 
-        getMainPanel().add(getInternalScrollPanel());
-        getMainPanel().add(getDomElementContainer());
-
-        add(getMainPanel());
+    void setupRootPanel() {
+        getRootPanel().add(getDomElementContainer());
+        getRootPanel().add(getScrollPanel());
     }
 
     void setupScrollHandlers() {
@@ -142,21 +156,40 @@ public class GridLienzoPanel extends FocusPanel implements RequiresResize,
 
     @Override
     public void updatePanelSize() {
-        final Element e = getElement().getParentElement();
-        final int width = e.getOffsetWidth();
-        final int height = e.getOffsetHeight();
+        final Element parentElement = getElement().getParentElement();
+        final Integer width = parentElement.getOffsetWidth();
+        final Integer height = parentElement.getOffsetHeight();
 
         if (width > 0 && height > 0) {
-            final Integer scrollbarWidth = getGridLienzoScrollHandler().scrollbarWidth();
-            final Integer scrollbarHeight = getGridLienzoScrollHandler().scrollbarHeight();
-
-            getDomElementContainer().setPixelSize(width - scrollbarWidth,
-                                                  height - scrollbarHeight);
-            getLienzoPanel().setPixelSize(width - scrollbarWidth,
-                                          height - scrollbarHeight);
-            getMainPanel().setPixelSize(width,
-                                        height);
+            updatePanelSize(width,
+                            height);
         }
+    }
+
+    @Override
+    public void updatePanelSize(final Integer width,
+                                final Integer height) {
+        updateScrollPanelSize(width,
+                              height);
+        updateInternalPanelsSizes(width,
+                                  height);
+    }
+
+    private void updateInternalPanelsSizes(final int width,
+                                           final int height) {
+        final Integer scrollbarWidth = getGridLienzoScrollHandler().scrollbarWidth();
+        final Integer scrollbarHeight = getGridLienzoScrollHandler().scrollbarHeight();
+
+        getDomElementContainer().setPixelSize(width - scrollbarWidth,
+                                              height - scrollbarHeight);
+        getLienzoPanel().setPixelSize(width - scrollbarWidth,
+                                      height - scrollbarHeight);
+    }
+
+    private void updateScrollPanelSize(final int width,
+                                       final int height) {
+        getScrollPanel().setPixelSize(width,
+                                      height);
     }
 
     @Override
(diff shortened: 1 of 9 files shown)


uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/layer/impl/GridLienzoPanel.java
```diff
@@ -147,18 +161,36 @@ public void updatePanelSize() {
...
+            updatePanelSize(width,
+                            height);
         }
     }
 
+    private void updatePanelSize(final int width,
+                                 final int height) {
+        updateScrollPanelSize(width,
+                              height);
+        updateInternalPanelsSize(width,
```
manstis: If I was being really picky I'd say this should be ```updateInternalPanelsSizes(..)``` ("size**s**").

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/scrollbars/GridLienzoScrollBounds.java
```diff
@@ -114,6 +114,7 @@ private double minValue(final List<Double> boundsValues) {
 
     private void addExtraBounds(final List<Double> bounds,
                                 final Function<Bounds, Double> function) {
+
```
manstis: Did you want to remove the blank line to make one less file in the PR?

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/test/java/org/uberfire/ext/wires/core/grids/client/widget/scrollbars/GridLienzoScrollBoundsTest.java
```diff
@@ -47,6 +47,7 @@
 
     @Before
     public void setUp() {
+
```
manstis: Did you want to remove the blank line to remove this file from the PR?

manstis:
@karreiro I've also noticed that all my mouse handlers don't work anymore too with this change :-(

For example, when you double-click on the cyan "Literal Expression" cell you should get a ```TextBox``` overlay the cell for editing it's value. Also, with the "Context" editor you should be able to adjust the column sizes by hovering over a column grid line in the body. This too does not work. I suspect the "Scroll Panel" is on top of the Lienzo Canvas and hence events are not propogating through? But only a guess.

karreiro:
Hi @manstis, thank you for reporting the problem. Everything is solved now.

I had to add two event handlers  (`setupContextSwitcher`) to switch between the canvas context and the scroll handler context. For now, this is the best that we can do.

However, in the future, I'll be able to remove everything related to the `setupContextSwitcher` and change the CSS a little bit to use *only* the `position: sticky` (https://developer.mozilla.org/en-US/docs/Web/CSS/position), this will make everything simpler. But it's not ready for production yet. 

Thanks again.

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/layer/impl/GridLienzoPanel.java
```diff
@@ -17,6 +17,7 @@
 
 import com.ait.lienzo.client.core.shape.Viewport;
 import com.ait.lienzo.client.widget.LienzoPanel;
+import com.google.gwt.core.client.GWT;
```
jomarko: Unused import

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/scrollbars/GridLienzoScrollHandler.java
```diff
@@ -22,6 +22,10 @@
 import com.ait.lienzo.client.core.shape.Viewport;
 import com.ait.lienzo.client.core.types.Transform;
 import com.ait.lienzo.client.widget.LienzoPanel;
+import com.google.gwt.core.client.GWT;
```
jomarko: Unused import

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/scrollbars/GridLienzoScrollUI.java
```diff
@@ -16,6 +16,7 @@
 
 package org.uberfire.ext.wires.core.grids.client.widget.scrollbars;
 
+import com.google.gwt.core.client.GWT;
```
jomarko: Unused import

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/scrollbars/GridLienzoScrollHandler.java
```diff
@@ -55,14 +59,28 @@ public void init() {
         setupGridLienzoScrollStyle();
         setupScrollBarSynchronization();
         setupMouseDragSynchronization();
+        setupContextSwitcher();
+    }
+
+    void setupContextSwitcher() {
+        getDomElementContainer().addDomHandler(disablePointerEvents(), MouseWheelEvent.getType());
```
manstis: Just curious why doesn't this use ```MouseMoveEvent``` (like that for enabling events on the _panel_)?

karreiro: @manstis Since the canvas parent DIV is `position: absolute` now, we want to disable `DomElementContainer` pointer events when the mouse wheel is rotated. This way, we enable scrollbars in the `ScrollPanel` (which would be blocked by `DomElementContainer` DIV).

karreiro:
Thank you for the review @jomarko and @manstis :-)

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/scrollbars/GridLienzoScrollHandler.java
manstis: @karreiro Heh. I wish I understood what you wrote! 

It all seemed to work "as *a User*" would expect so I'm happy!

karreiro: :-(

I'm a terrible explainer.

but, look:
image


Without the `pointer-events` thing, if the user scrolls with the cursor inside of the `DomElementContainer`, the scrollbars from the `DomElementContainer` DIV would be activated (and the `ScrollPanel` would continue static).

However, we want to enable events on `DomElementContainer` when the user interacts with the canvas, _and_ we want to enable `ScrollPanel` events when the user rotates the mouse wheel.

That's the why this _switcher_ is needed.

manstis: Love it!

I understand the use of ```pointer-events``` (must have taken some googling!! ;-)) and ```MouseMoveHandler``` on the ```DomElementContainer``` just not why it's only the ```MouseWheelHandler``` for the scroll panel.. hence I asked why not ```MouseMoveHandler``` too.

karreiro: @manstis Most of the time we want the canvas (`DomElementContainer`) enabled, everything happens there.

However, we want to ignore canvas events **only** when the user rotates the mouse wheel. Only when the user rotates the wheel, because when the user is moving the cursor, clicking, dragging, he/she intends to interact with the canvas, however when the user is rotating the mouse wheel, we want to enable the `ScrollPanel` and ignore the `canvas` events.

manstis: Yay. Now I understand :-)

PR #828, opened 2017-09-14
GUVNOR-3479 : [GSS-RFE] Add support for missing keywords like "not matches" condition in Guided Rule

Rikkola:
https://issues.jboss.org/browse/GUVNOR-3479

This updates a list. Do we want to test the list content? 
Right now we don't do that.


Rikkola:
@manstis Yep. Was surprised how it worked with no additional changes. Opening, reopening, compile and validate works.

I'll add some checks for all the lists in the class. Not sure how good they end up being. If you refactor the lists then you probably also edit the lists, but I'll try to make something.

Rikkola:
Alrighty, threw some tests in.

PR #835, opened 2017-09-19
GUVNOR-2825: [Guided Decision Table] Hints are empty

markcoble:
See https://issues.jboss.org/browse/GUVNOR-2825

Setting title/content via super class constructor and calling recreate method.  The init() method of Popover class was never being called from InfoPopup so help messages were never displayed.

If the formatting is not liked don't blame me!  It is the latest Eclipse formatter.


markcoble:
@manstis agreed.  Just wanted to put this out for review first.

manstis:
@markcoble Will approve when the Unit Test(s) are provided. Thanks.

markcoble:
@manstis unit tests added.

manstis:
@markcoble Please let me know when green.

markcoble:
@manstis all green.

PR #838, opened 2017-09-19
GUVNOR-3487: Multi-select combo box in guided decision table only has a height of a single row

manstis:
See https://issues.jboss.org/browse/GUVNOR-3487


manstis:
Part of an ensemble:

- https://github.com/AppFormer/uberfire/pull/838
- https://github.com/kiegroup/drools-wb/pull/624

manstis:
@jomarko For your attention, please!

PR #843, opened 2017-09-22
GUVNOR-3511: Add possibility to change prompt and comment visibility

jomarko:
@manstis This is  ensemble with kiegroup/drools-wb#629


jomarko:
@manstis @karreiro replaced by #844 due to different branch names across the ensemble.

PR #845, opened 2017-09-26
GUVNOR-3344: [Guided Decision Table] Editing BRL columns does not work

karreiro:
See https://issues.jboss.org/browse/GUVNOR-3344

Before:
before

After:
new-1
new-2

### DIFF
--- a/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/model/impl/BaseGridData.java
+++ b/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/model/impl/BaseGridData.java
@@ -20,6 +20,8 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.OptionalInt;
+import java.util.stream.IntStream;
 
 import org.uberfire.commons.validation.PortablePreconditions;
 import org.uberfire.ext.wires.core.grids.client.model.GridCell;
@@ -90,7 +92,8 @@ public class BaseGridData implements GridData {
         if (column.getColumnRenderer() instanceof HasDOMElementResources) {
             ((HasDOMElementResources) column.getColumnRenderer()).destroyResources();
         }
-        columns.remove(column);
+
+        removeColumn(column);
 
         //Destroy column data
         for (GridRow row : rows) {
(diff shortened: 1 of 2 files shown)


karreiro:
@manstis @jomarko Could you please take a look?

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/model/impl/BaseGridData.java
```diff
@@ -108,6 +109,19 @@ public void deleteColumn(final GridColumn<?> column) {
         selectionsManager.onDeleteColumn(index);
     }
 
+    private void removeColumn(final GridColumn<?> column) {
+
+        for (int index = 0; index < columns.size(); index++) {
```
jomarko: This could be shortened like shown below, I think:
```java
IntStream.range(0, columns.size())
                .filter(i -> column == columns.get(i))
                .findFirst()
                .ifPresent(i -> columns.remove(i));
```

karreiro:
Thanks for the suggestion, @jomarko. Changes applied.

---

Thanks for the review, @manstis. I slightly changed the PR, adding the `columns.remove(obj)` as a fallback.

I don't think that we should change the `equals` method on `BaseGridColumn`. The scenario that we have is:

1) a table with the following columns:
```
- #
- Description
- My BRL Column - age
```
2) and then, the user decides to update the table by adding a new child column (jobPosition). Now we need to add a new "My BRL Column" with extra fields (a) and remove the old one (b):

a) adding extra fields
```
- #
- Description
- [new] My BRL Column - age
- [new] My BRL Column - jobPosition
- [old] My BRL Column - age
```
b) removing the old one
```
- #
- Description
- [new] My BRL Column - age
- [new] My BRL Column - jobPosition
```
The point is that `[old] My BRL Column - age` and `[new] My BRL Column - age` are equal. However we *need* to remove the old reference, so we need to remove a particular instance.

Since and I'm changing the `BaseGridData.deleteColumn`, sounds fair enough to keep the fallback.

karreiro:
@manstis no problem :-) ...and, yes, another possible fix is to call the delete method before the "adds" (+ some extra work).

However, by changing the `BaseGridData.delete` method, it can be called after and before the "adds", it works fine anyway. So, it sounds like good approach :-)

manstis:
@karreiro I am convinced. Good job (in the analysis, fix and convincing me!)

PR #847, opened 2017-09-28
GUVNOR-3505: Add detailed pipeline error information for Guvnor ALA UI

wmedvede:
- Adds the ability of showing the ErrorPopup with an optional detail message, and the ability of using the PatternFly or Standard style


uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/widgets/ErrorPopupView.java
```diff
@@ -0,0 +1,215 @@
+/*
+ * Copyright ${year} Red Hat, Inc. and/or its affiliates.
```
manstis: 2017

wmedvede:
Hello @cristianonicolai would you mind review this PR?
Includes some improvements in the ErrorPopup that we need for showing detailed pipeline errors. The by default behavior remains the same.

Thanks.

wmedvede:
Hi @ederign   If you could take a look at this it'll be great since this small improvement is needed for the 7.3. So if merged I can then cherry-pick to the UF branch used by the 7.3 release.
Thanks.

PR #861, opened 2017-10-06
GUVNOR-3200: [Guided Decision Table] Access control to prohibit business users from modifying columns

karreiro:
See https://issues.jboss.org/browse/GUVNOR-3200

---

permissoes

---

Part of an ensemble:
- https://github.com/AppFormer/uberfire/pull/861
- https://github.com/kiegroup/kie-wb-common/pull/1190
- https://github.com/kiegroup/drools-wb/pull/636
- https://github.com/kiegroup/kie-wb-distributions/pull/628
- https://github.com/kiegroup/kie-docs/pull/402

### DIFF
--- a/uberfire-security/uberfire-security-api/src/main/java/org/uberfire/security/impl/authz/DefaultAuthorizationEntry.java
+++ b/uberfire-security/uberfire-security-api/src/main/java/org/uberfire/security/impl/authz/DefaultAuthorizationEntry.java
@@ -23,10 +23,11 @@ import org.uberfire.security.authz.PermissionCollection;
 @Portable
 public class DefaultAuthorizationEntry {
 
+    public static final int DEFAULT_PRIORITY = -100;
     private String description = null;
     private Role role = null;
     private Group group = null;
-    private int priority = 0;
+    private int priority = DEFAULT_PRIORITY;
     private String homePerspective = null;
     private PermissionCollection permissions = new DefaultPermissionCollection();
 
--- a/uberfire-extensions/uberfire-security/uberfire-security-management/uberfire-widgets-security-management/src/test/java/org/uberfire/ext/security/management/client/widgets/management/editor/acl/ACLSettingsTest.java
+++ b/uberfire-extensions/uberfire-security/uberfire-security-management/uberfire-widgets-security-management/src/test/java/org/uberfire/ext/security/management/client/widgets/management/editor/acl/ACLSettingsTest.java
@@ -35,6 +35,7 @@ import org.uberfire.security.impl.authz.DefaultPermissionManager;
 import org.uberfire.security.impl.authz.DefaultPermissionTypeRegistry;
 
 import static org.mockito.Mockito.*;
+import static org.uberfire.security.impl.authz.DefaultAuthorizationEntry.DEFAULT_PRIORITY;
 
 @RunWith(MockitoJUnitRunner.class)
 public class ACLSettingsTest {
@@ -69,7 +70,7 @@ public class ACLSettingsTest {
 
         permissionManager.setAuthorizationPolicy(permissionManager.newAuthorizationPolicy()
                                                          .role("admin").home("HomeAdmin").priority(10)
-                                                         .group("group1").home("HomeGroup1").priority(0)
+                                                         .group("group1").home("HomeGroup1").priority(DEFAULT_PRIORITY)
                                                          .build()
         );
 
@@ -87,7 +88,7 @@ public class ACLSettingsTest {
         when(homePerspectiveDropDown.getItemName("HomeAdmin")).thenReturn("HomeAdmin");
         when(homePerspectiveDropDown.getItemName("HomeGroup1")).thenReturn("HomeGroup1");
         when(priorityDropDown.getPriorityName(10)).thenReturn("High");
-        when(priorityDropDown.getPriorityName(0)).thenReturn("Normal");
+        when(priorityDropDown.getPriorityName(DEFAULT_PRIORITY)).thenReturn("Very Low");
     }
 
     @Test
@@ -125,7 +126,7 @@ public class ACLSettingsTest {
         verify(view).setPrioritySelector(any());
         verify(view).setHomePerspectiveName("HomeGroup1");
         verify(view).setHomePerspectiveTitle("HomeGroup1");
-        verify(view).setPriorityName("Normal");
+        verify(view).setPriorityName("Very Low");
     }
 
     @Test
@@ -137,7 +138,7 @@ public class ACLSettingsTest {
         verify(view).setHomePerspectiveSelector(any());
         verify(view).setPrioritySelector(any());
         verify(homePerspectiveDropDown).setSelectedPerspective("HomeGroup1");
-        verify(priorityDropDown).setSelectedPriority(0);
+        verify(priorityDropDown).setSelectedPriority(DEFAULT_PRIORITY);
     }
 
     @Test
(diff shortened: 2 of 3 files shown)


karreiro:
@manstis Thank you for the suggestion. I decided to set the default priority to `Integer.MIN_VALUE` as you recommended.

Previously, I misunderstood the real issue (curiously, I could apply the same fix to solve).

The permissions are being replaced here (because every user has a default group with his name with default permissions):
https://github.com/AppFormer/uberfire/blob/f8f3737a854dc64f1075b076514c8630bf3da5f2/uberfire-security/uberfire-security-api/src/main/java/org/uberfire/security/impl/authz/DefaultPermissionManager.java#L297

manstis:
@dgutierr Could you please review too.

@karreiro and I have experienced _unexpected_ behaviour with default permissions and User/Group/Role overrides. In such scenarios the behaviour (of resolved permission) can be resolved by making the Group/Role priority higher (than _normal_). @karreiro and I thought this was the best solution; to make default permissions have the lowest priority so any User specific setting overrides correctly (without having to alter the Group/Role's priority). 

@jomarko This could explain what you found on https://github.com/kiegroup/kie-wb-distributions/pull/631#issuecomment-334686138 (I'll comment that PR too about your other comments).

jomarko:
Please give me a chance to review, I am looking right now.

uberfire-security/uberfire-security-api/src/main/java/org/uberfire/security/impl/authz/DefaultAuthorizationEntry.java
```diff
@@ -26,7 +26,7 @@
     private String description = null;
     private Role role = null;
     private Group group = null;
-    private int priority = 0;
+    private int priority = Integer.MIN_VALUE;
```
dgutierr: TBH I wouldn't change the default priority to MIN_VALUE as this might have side effects in the permission resolution logic. See:

https://github.com/AppFormer/uberfire/blob/master/uberfire-security/uberfire-security-api/src/main/java/org/uberfire/security/impl/authz/DefaultPermissionManager.java#L285
https://github.com/AppFormer/uberfire/blob/master/uberfire-security/uberfire-security-api/src/main/java/org/uberfire/security/impl/authz/DefaultAuthorizationPolicy.java#L208

If see two alternatives to this:

1.- Set "private int priority = -100;" for instance  (or whatever is lower than -10 according to https://github.com/AppFormer/uberfire/blob/master/uberfire-extensions/uberfire-security/uberfire-security-management/uberfire-widgets-security-management/src/main/java/org/uberfire/ext/security/management/client/widgets/management/editor/acl/PriorityDropDown.java#L109)

2.- Undo this change => "private int priority = 0;" and define the priority of each role / group at the security-policy-properties files. For example => https://github.com/AppFormer/uberfire/blob/master/uberfire-showcase/uberfire-webapp/src/main/resources/security-policy.properties#L26  I Personally I like this 2nd option as it translates the priority resolution to app level.

manstis: @dgutierr Can you please point me to where the discrete values used by the UI are defined ("very high", "high", "normal", "low", "very low")? Can I also assume _default_ is 0 (if @karreiro change is undone?) If we change a role to have a priority other than one used by the UI I assume there's a risk things will break.

dgutierr: @manstis Do you mean https://github.com/AppFormer/uberfire/blob/master/uberfire-extensions/uberfire-security/uberfire-security-management/uberfire-widgets-security-management/src/main/java/org/uberfire/ext/security/management/client/widgets/management/editor/acl/PriorityDropDown.java#L93 ?

karreiro: @dgutierr Thank you for the review, David. :-) I've preferred to apply the first alternative. I think that having the default permissions with a lower priority (by default) makes sense.

@manstis I think that the UI is based on ranges:
https://github.com/karreiro/uberfire/blob/f8f3737a854dc64f1075b076514c8630bf3da5f2/uberfire-extensions/uberfire-security/uberfire-security-management/uberfire-widgets-security-management/src/main/java/org/uberfire/ext/security/management/client/widgets/management/editor/acl/PriorityDropDown.java#L93-L107

but.. let's wait for @dgutierr comments.

manstis: Thanks @dgutierr @karreiro exactly what I was looking for (and would you believe it, linked by @dgutierr in his original Option 1.. if only I'd read it fully)... as for which option is best.. can we add ```default.priority==-100``` in the configuration file?

dgutierr: @manstis I didn't try it tbh, but I think it will work. Nice suggestion btw.

manstis: @dgutierr @karreiro It does not work.. ```priority``` is expected to follow a [group or role](https://github.com/AppFormer/uberfire/blob/master/uberfire-security/uberfire-security-api/src/main/java/org/uberfire/security/impl/authz/AuthorizationPolicyBuilder.java#L157). I tried deploying a WAR with the proposed setting but it failed.

@karreiro Personally I quite like @dgutierr second option.. defining the priority of other roles with higher priorities than default is exactly the same as defining default to have lower priorities than the others...  but then I agree with you that you'd simply expect specific permissions to override the defaults. Changing properties files works for our "out of the box" roles; but would not work for a custom group added by the customer (in EAP/WF etc). 

Oh.. I just can't decide.. go with your head @karreiro :-)

jomarko:
I see there is quite a heavy development around this issue. I wouldn't like to block you. I wont be able to finish my review today. So if you need, feel free to merge this one without my ack. I can still do some post merge testing.

uberfire-security/uberfire-security-api/src/main/java/org/uberfire/security/impl/authz/DefaultAuthorizationEntry.java
karreiro: @manstis Hahaha... ok so. Considering all the comments...

I still like the **first alternative** that David proposed :-)

karreiro:
@manstis and @dgutierr Changes applied. Could you please take a look? :-)

uberfire-security/uberfire-security-api/src/main/java/org/uberfire/security/impl/authz/DefaultAuthorizationEntry.java
dgutierr: @manstis @karreiro Looking thoroughly at the code,  I realized a default priority setting will require some extra development. Not a crazy idea whatsoever.  I can open a JIRA in case you vote for it.

dgutierr:
@karreiro The changes look good :-)

uberfire-security/uberfire-security-api/src/main/java/org/uberfire/security/impl/authz/DefaultAuthorizationEntry.java
manstis: @dgutierr @karreiro needs this change in master (and 7.3.x) for LA... so your improvement will need to follow in 7.5 or 8.. Thanks for your insights on this PR.

manstis: @dgutierr Can you please approve this?

PR #869, opened 2017-10-11
AF-618: [Perspective Editor] Disable the "Tags" button

dgutierr:
The Tags button will remain hidden by default. Only the uberfire-webapp and uberfire-wires-webapp keep it visible.

@ederign Can you review/approve please?


jhrcek:
@dgutierr @ederign I just noticed that this change makes AppsPerspective useless - now that we can't associate Tags with Perspectives (now called Pages) there's no way how to make something appear in the Apps perspective. Or am I missing something? Should I file a jira for that?

dgutierr:
@jhrcek @ederign I also agree the App perspective is no longer needed. However, the workbench's menu entry "Business Dashboards" is pointing to the Apps perspective, which I think is an error as it should open the CMS one?. Also, I'm wondering why an additional "Dashboard" entry exists in the  menu. @paulovmr Can you please shed some light on this?

paulovmr:
@dgutierr With this change I also agree that the `AppsPerspective` becomes useless. The idea of the "Design > Dashboards" entry is to take the user to `ContentManagerPerspective` to design new dashboards. The idea of "Track > Business Dashboards" is to the user see those designed dashboards running. So in my view, the "Design > Dashboards" should remain pointing to `ContentManagerPerspective` and the "Track > Business Dashboards" should be removed.

@porcelli WDYT?

porcelli:
@paulovmr go for that!

dgutierr:
@paulovmr I do agree!  Please do not forget to mention such changes in the release notes, to avoid any confusion.

PR #872, opened 2017-10-12
AF-634: fix for SSH issue not propagating to the rest of the cluster.

porcelli:
Now the clusterService injection to JGitFileSystemProvider is simplified
(in the end we never managed to isolate cluster per repository anyway).

### DIFF
--- a/uberfire-io/src/main/java/org/uberfire/io/impl/cluster/helix/ClusterServiceHelix.java
+++ b/uberfire-io/src/main/java/org/uberfire/io/impl/cluster/helix/ClusterServiceHelix.java
@@ -16,6 +16,7 @@
 
 package org.uberfire.io.impl.cluster.helix;
 
+import java.net.URI;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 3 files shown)


ederign:
@porcelli thanks for this. I only suggest surrounding unlock on L586 on a try/catch/finally block.

uberfire-io/src/main/java/org/uberfire/io/impl/cluster/helix/ClusterServiceHelix.java
```diff
@@ -70,6 +74,10 @@ public ClusterServiceHelix( final String clusterName,
         this.participantManager = getZkHelixManager( clusterName, zkAddress, instanceName );
         PriorityDisposableRegistry.register( this );
         start();
+        final FileSystemProvider provider = FileSystemProviders.resolveProvider(URI.create("git://test"));
```
manstis: It might have been nicer to make ```JGitFileSystemProvider.SCHEME``` public and simply use ```FileSystemProviders.getProvider( JGitFileSystemProvider.SCHEME )```.. but this does the same job!

uberfire-nio2-backport/uberfire-nio2-impls/uberfire-nio2-jgit/src/main/java/org/uberfire/java/nio/fs/jgit/JGitFileSystemProvider.java
```diff
@@ -316,7 +316,11 @@ private void setupProxyAuthentication( final String httpProxyUser,
                                                           httpsProxyPassword ) );
     }
 
-    public void onCloseFileSystem( final JGitFileSystem fileSystem ) {
+    public void setupClusterService(final ClusterService clusterService) {
```
manstis: It's not as such a "set up" more of simply a "set".

PR #891, opened 2017-10-30
AF-890: avoid index respositories that aren't really needed

porcelli:


uberfire-structure/uberfire-structure-backend/src/test/java/org/guvnor/structure/backend/repositories/git/GitRepositoryFactoryHelperNoIndexTest.java
```diff
@@ -0,0 +1,178 @@
+/*
+ * Copyright 2014 Red Hat, Inc. and/or its affiliates.
```
manstis: 2017..

PR #903, opened 2017-11-09
RHBRMS-3022: Cannot disable "Delete" button for Analyst role in business central

karreiro:
See https://issues.jboss.org/browse/RHBRMS-3022

Before:
before

After:
after

---

Menus from other editors were interfering with permissions to enable/disable delete (and other buttons) in the Project Editor.

---

Part of an ensemble:
- https://github.com/AppFormer/uberfire/pull/903
- https://github.com/kiegroup/kie-wb-common/pull/1284

### DIFF
--- a/uberfire-api/src/main/java/org/uberfire/workbench/model/menu/impl/MenuBuilderImpl.java
+++ b/uberfire-api/src/main/java/org/uberfire/workbench/model/menu/impl/MenuBuilderImpl.java
@@ -26,6 +26,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.Stack;
 
+import com.google.common.base.Joiner;
 import org.uberfire.mvp.Command;
 import org.uberfire.mvp.PlaceRequest;
 import org.uberfire.mvp.impl.DefaultPlaceRequest;
@@ -63,7 +64,7 @@ public final class MenuBuilderImpl
     final Stack<MenuFactory.CustomMenuBuilder> context = new Stack<MenuFactory.CustomMenuBuilder>();
     int order = 0;
 
-    private static class CurrentContext implements MenuFactory.CustomMenuBuilder {
+    static class CurrentContext implements MenuFactory.CustomMenuBuilder {
 
         MenuItem menu = null;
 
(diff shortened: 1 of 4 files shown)


karreiro:
Jenkins please retest this.

uberfire-api/src/main/java/org/uberfire/workbench/model/menu/impl/MenuBuilderImpl.java
```diff
@@ -372,11 +371,10 @@ public void addEnabledStateChangeListener( final EnabledStateChangeListener list
...
                 @Override
                 public String getSignatureId() {
-                    if ( contributionPoint != null ) {
-                        return getClass().getName() + "#" + contributionPoint + "#" + caption;
 
-                    }
-                    return getClass().getName() + "#" + caption;
+                    final String[] signatureParts = { getClass().getName(), contributionPoint, namespace, caption };
+
+                    return Joiner.on( '#' ).skipNulls().join( signatureParts );
```
jomarko: I see this is for 6.5.x branch, if here is not possible to use java stream I am fine with Using this `Joiner` otherwise I would ask you for something like:
```java
return Stream.of(signatureParts)
               .filter(s -> s != null)
               .collect(Collectors.joining("#"));
```

uberfire-api/src/main/java/org/uberfire/workbench/model/menu/impl/MenuBuilderImpl.java
```diff
@@ -545,6 +543,12 @@ public MenuBuilderImpl withRoles( final String... roles ) {
     }
 
     @Override
+    public MenuBuilderImpl withNamespace(final String namespace) {
+        ( (CurrentContext) context.peek() ).namespace = checkNotEmpty( "contributionPoint", namespace );
```
manstis: This should probably be ```checkNotEmpty( "namespace", namespace )``` ?

uberfire-api/src/main/java/org/uberfire/workbench/model/menu/impl/MenuBuilderImpl.java
karreiro: Yes. But, the `6.5.x` branch supports JDK7 for all modules, so the `Joiner` is needed.

karreiro:
@manstis @jomarko Changes applied.

Thanks for the review.

PR #906, opened 2017-11-09
Support elemental2 in workbench views + upgrade Errai snapshot version.

mbarkley:


mbarkley:
Related:
https://github.com/errai/errai/pull/301
https://github.com/AppFormer/uberfire/pull/906
https://github.com/dashbuilder/dashbuilder/pull/401
https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/605

mbarkley:
@ederign @adrielparedes Can you please review this?

This change adds an UberElemental interface that is like UberElement but can be used for views composed of elemental2 element wrappers and the new corresponding org.jboss.errai.common.client.api.elemental2.IsElement interface.

Along the way I had to update the Errai snapshot version, and I consolidated the init methods from UberView and UberElement into a single HasPresenter interface, which makes the templates a bit simpler now that there are three possible view kinds.

mbarkley:
Jenkins, please execute full downstream build.

mbarkley:
The only downstream test failure is in a module that does not depend on uberfire, so almost certainly unrelated.

mbarkley:
Jenkins, execute full downstream build.

PR #910, opened 2017-11-10
Library UX/UI improvements

paulovmr:
AF-636: UX/UI improvements on spaces (teams) screen
AF-892: UX/UI improvements on projects screen

### DIFF
--- a/uberfire-extensions/uberfire-social-activities/uberfire-social-activities-client/src/test/java/org/ext/uberfire/social/activities/client/SocialEntryPointTest.java
+++ b/uberfire-extensions/uberfire-social-activities/uberfire-social-activities-client/src/test/java/org/ext/uberfire/social/activities/client/SocialEntryPointTest.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.ext.uberfire.social.activities.client;
+
+import org.ext.uberfire.social.activities.service.SocialUserRepositoryAPI;
+import org.jboss.errai.common.client.api.Caller;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+import org.uberfire.mocks.CallerMock;
+import org.uberfire.mocks.SessionInfoMock;
+import org.uberfire.rpc.SessionInfo;
+
+import static org.mockito.Mockito.verify;
+
+@RunWith(MockitoJUnitRunner.class)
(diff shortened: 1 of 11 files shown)


paulovmr:
Part of an ensemble:
* https://github.com/AppFormer/uberfire/pull/910
* https://github.com/kiegroup/kie-wb-common/pull/1289

uberfire-structure/uberfire-structure-client/src/test/java/org/guvnor/structure/organizationalunit/OrganizationalUnitServiceCallerMock.java
```diff
@@ -25,13 +25,13 @@
 
 /**
  * Utility class for client side testing.
- * <p>
+ * <p/>
```
tomasdavidorg: I think p is pair tag. I think it is possible to use only the starting tag `<p>` or the second style is with the ending tag `</p>` at the end of the paragraph.

uberfire-extensions/uberfire-social-activities/uberfire-social-activities-client/src/test/java/org/ext/uberfire/social/activities/client/SocialEntryPointTest.java
```diff
@@ -0,0 +1,57 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+import org.uberfire.mocks.CallerMock;
+import org.uberfire.mocks.SessionInfoMock;
+import org.uberfire.rpc.SessionInfo;
+
+import static org.mockito.Mockito.*;
```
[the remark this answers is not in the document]
paulovmr: @tomasdavidorg Fixed.

uberfire-structure/uberfire-structure-client/src/test/java/org/guvnor/structure/organizationalunit/OrganizationalUnitServiceCallerMock.java
paulovmr: @tomasdavidorg Fixed.

uberfire-structure/uberfire-structure-backend/src/main/java/org/guvnor/structure/backend/organizationalunit/OrganizationalUnitServiceImpl.java
```diff
@@ -194,6 +224,17 @@ public OrganizationalUnit createOrganizationalUnit(final String name,
     public OrganizationalUnit updateOrganizationalUnit(final String name,
                                                        final String owner,
                                                        final String defaultGroupId) {
+        return updateOrganizationalUnit(name,
+                                        owner,
+                                        defaultGroupId,
+                                        null);
```
jhrcek: (optional) wouldn't it make sense to pass some empty collection instead of null to this method here? I've seen a lot of bugs caused by NPEs recently and I prefer using empty collection values instead of null whenever it makes sense.. (Not to mention that debugging those "Unexpected exception occurred: TypeError: d.f was Undefined" is extremely time consuming).

You're already passing empty array list above in `createOrganizationalUnit` so in my opinion it would make sense to do this consistently throughout this PR.

paulovmr: Not necessary in this case, the method being called makes the check to avoid a NPE. When the OU is being updated, if the passed list is null, the contributors property is not updated, which is the idea.

ederign:
note: before merging this, please wait for @barboras7 "go" on https://github.com/kiegroup/kie-wb-common/pull/1289  and my smoke test tomorrow morning.

PR #911, opened 2017-11-13
AF-617: [Layout Editor] Component palette reorganization

dgutierr:
The interface PerspectiveEditorComponentGroup must be implemented by every group listed in the components palette. Every group implementation is in charge of looking up its drag components. Two groups have been declared by default: "Core" and "Navigation" which can be extended with further group implementations if need be. See screenshots:

navigationgroup
coregroup

@ederign @pefernan @kkufova Can you take a look please?


kkufova:
@dgutierr, I like these changes a lot. I've noticed though that for some reason, sometimes both of the new panels can be expanded and sometimes only one of them -- which of these is the correct behavior? Let me know if you want me to send you the exact steps to reproduce this (I have 3 custom pages, if I create a new one, both can be expanded, but if I close the newest one and open an older one, only one panel can be expanded. I am not sure though if this would work for you as well.)

Also, just a very small detail: do we need the word "component" in the name of each item? It feels kinda redundant.

ederign:
jenkins retest this

dgutierr:
@kkufova Thanks for the feedback!

I made a couple of changes:

1- Both panels can be expanded at the same time.
2- I removed the word "Component" all across the palette as you suggested.

Please have a look :-)

PR #931, opened 2018-06-04
[RHBPMS-5142] User with role "user" can get edit access to some assets

elguardian:
workaround for user bypassing security. This avoids the possibility
of rendering parts of the UI without the security being checked

### DIFF
--- a/uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/mvp/PlaceHistoryHandler.java
+++ b/uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/mvp/PlaceHistoryHandler.java
@@ -134,8 +134,8 @@ public class PlaceHistoryHandler {
                 historian.addValueChangeHandler( new ValueChangeHandler<String>() {
                     @Override
                     public void onValueChange( ValueChangeEvent<String> event ) {
-                        String token = event.getValue();
-                        handleHistoryToken( token );
+//                        String token = event.getValue();
+//                        handleHistoryToken( token );
                     }
                 } );
 


elguardian:
same as 
https://github.com/kiegroup/appformer/blob/master/uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/mvp/PlaceHistoryHandler.java#L72

porcelli:
@elguardian how disabling history, fix the security issue?

mareknovotny:
@porcelli we are waiting for this for CP and if it is not correct let us know and to not wait for it.

porcelli:
@mareknovotny @elguardian i'm still trying to understand the relationship between back button and the security. with that said... it doesn't quite look correct, but I'll wait for @elguardian explain why back button is the issue.

mareknovotny:
ok, thanks @porcelli

elguardian:
@porcelli 
https://issues.jboss.org/browse/RHBPMS-5142?focusedCommentId=13571971&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13571971

https://issues.jboss.org/browse/RHBPMS-5142?focusedCommentId=13573390&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13573390

elguardian:
@porcelli (even if it didin't make the cut.  
let me rephrase. The trigger reacts against user input (copy&paste) of the uri, causing "a go to" and rendering that part (the editor in this case). Because there is no restriction associated it renders that part (by default is ok)
https://github.com/AppFormer/uberfire/blob/0.9.x/uberfire-security/uberfire-security-api/src/main/java/org/uberfire/security/impl/authz/RuntimeResourceManager.java#L88.

Disabling this causes this unwanted behaviour not to happen. This is not a proper fix. I warned @mareknovotny  about that.

jomarko:
@elguardian could we please remove the commented code and rather provide comments why the handler is empty? Also this PR needs some unit test I think.

elguardian:
@jomarko Are you ok then with the PR ? If it is ok I will performe those changes you were asking.

jomarko:
@elguardian Hi, I still think this can not be merged as we should add test into the class `PlaceHistoryHandlerTest`. The new tets should assert the behavior of the changed handler. Similar tests can be found here: [ColumnsPagePresenterTest#testRuleSelectorWidget](https://github.com/kiegroup/drools-wb/blob/master/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/src/test/java/org/drools/workbench/screens/guided/dtable/client/editor/page/ColumnsPagePresenterTest.java)

ederign (changes requested):
I don't think this is the right solution for this. We are just killing the messenger (the history at the url) and not fixing the issue.

elguardian:
@ederign Problem here is not about fixing the behaviour but how this works. 
Try to change the behavour of this would burst into a more complex issue probably leading to something more related to security. As I stated before in the jira this has to do how the security is calculated. It is not as fine grained as granting permissions to render a part or not (by default everthing that is not restricted is ok to go). This is happening when the app reacts to uri fragment change.

Probably because in upstream is commented we cannot reproduce the issue there
https://github.com/kiegroup/appformer/blob/master/uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/mvp/PlaceHistoryHandler.java#L73

But more likely if the security part has not change from the version used by 6.5.x, this would happen as well.
https://github.com/AppFormer/uberfire/blob/0.9.x/uberfire-security/uberfire-security-api/src/main/java/org/uberfire/security/impl/authz/RuntimeResourceManager.java#L81

Unfortunatly I don't see a way to get the activity  at the PlaceHistoryHandler  listener, otherwise I would try to set that and probably the security would be calculated properly.

Not really sure if this makes sense to you.

ederign:
@elguardian the idea behind your PR is to disable history on the browser right?

But how this fix the security issue if the user can type the URL and get the same behavior??

elguardian:
@ederign
I am ok if you want to close this but regarding this:
 
>> But how this fix the security issue if the user can type the URL and get the same behavior??

Do you mind to point me out where in the code (uberfire) is reacting against URI change ?

From my point of view there are two ways to approach properly the issue:
1) fix how security is calculated (I don't think this is feasible) for a CP
2) try to fix the new place request and provide correct input to the goto function https://github.com/kiegroup/appformer/blob/master/uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/mvp/PlaceHistoryHandler.java#L119 (I don't see a way to do this eihter)

porcelli:
@elguardian if this is not a real fix, what's the point to just disable browser history? as pointed by @ederign this can be workarounded in no time.

@elguardian there're permissions on the open of a file, you need to address it in there

